PR #24, opened 2017-12-20
Add ARM Template to deploy Azure Resources

tomasr:
Adds an ARM template that deploys:
* App Service Plan
* Application Insights Resource
* Web App
  * AppInsights Site Extension
  * ASP.NET Core Site Extension
* Key Vault
  * Admin/AzureAD - Secrets


onovotny:
Thank you so much!

I just removed three pieces of configuration in favor of querying on startup - to make it easier for users.

`Location`, `ApplicationObjectId`, and `Domain` were removed.

I'll test this out this afternoon!

onovotny:
This looks really great, thanks for this.

Couple things I noticed:

1. The default ACL's of the key vault should probably include full permissions to the person executing the template (or at least all secret permissions). Right now if you go to the portal to look, you'd have to add an ACL for yourself.
2.  Is it possible to default the `hostingPlanName` and `keyVaultName` to be derived from the website name? Wondering if we can reduce parameters (or at least provide defaults if people don't want to provide them).

I moved the templates into VS resource project as that includes goodies for exploring the resources and easily deploying from VS. I also rebased off of master.

Is this an Add/Update operation? I.e., can it be run again with the same parameters and upsert entries?

I'm thinking that I can prompt the user to run this as part of the Install Utility where the utility already knows all of the parameters. The install utility can then also add the appropriate ReplyUrl since it'll know the value created.

What kind of error handling is normally in these template in case of name collisions? I.e., suppose someone picks a website name that's already taken? Or they pick an available website name but the derived keyvault name is either invalid or taken?

tomasr:
>The default ACL's of the key vault should probably include full permissions to the person executing the template (or at least all secret permissions). Right now if you go to the portal to look, you'd have to add an ACL for yourself.

No easy way to do this, as far as I know. Also only makes sense when you are deploying manually, but wouldn't make sense for an automated deployment.

> Is it possible to default the hostingPlanName and keyVaultName to be derived from the website name? Wondering if we can reduce parameters (or at least provide defaults if people don't want to provide them).

Done now.

> Is this an Add/Update operation? I.e., can it be run again with the same parameters and upsert entries?

That depends on whether you run the template in Incremental or Replace mode. But yeah, should be safe to redeploy to make minor changes.

> What kind of error handling is normally in these template in case of name collisions? I.e., suppose someone picks a website name that's already taken? Or they pick an available website name but the derived keyvault name is either invalid or taken?

Any resource that fails to create, will cause all other resources that depend on it to not be created. That said, I added some `uniqueString()` goodness to the App Service Plan and Key Vault names to minimize the chance they'll clash.

onovotny:
Thanks!!!

PR #29, opened 2018-02-11
Upgrade AzureSignTool.

vcsjones:
Minor performance improvement.

https://github.com/vcsjones/AzureSignTool/releases/tag/1.0.1


onovotny:
Cool, I assume it's just that one file, the rest of the binaries are the same?

vcsjones:
Yes - I copied all of them in to be safe and just the one showed up as a change in git.

PR #46, opened 2018-08-27
Support msix/msixbundle signing on Windows 10 1607/Windows Server 2016+

riverar:


vcsjones:
Neat! Presumably these same changes for the CLI tool would work too?

riverar:
@vcsjones Yup. Nothing really novel or new here, just glued together everything @onovotny had written but in the right order. He's upset 😆.

PR #58, opened 2018-11-25
Use public nuget package.

vcsjones:
This also accounts for some API surface changes.


vcsjones:
@onovotny not sure why build failed, I can't see results. Can you please take a look?

PR #69, opened 2019-05-07
Deploy Application Insights to resource group region

Aaronontheweb:
Value used to be hard coded to `East US` region. Now uses whatever was intended for the rest of the resource group.

### DIFF
--- a/arm/ArmDeploy/azuredeploy.json
+++ b/arm/ArmDeploy/azuredeploy.json
@@ -185,7 +185,7 @@
       "apiVersion": "2014-04-01",
       "name": "[variables('appInsightsName')]",
       "type": "Microsoft.Insights/components",
-      "location": "East US",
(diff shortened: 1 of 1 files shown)


onovotny:
Has App Insights been deployed to more regions? When I last checked, and it has been a little while, there were surprisingly few regions where it was available so it seemed safer to hard-code it.

Aaronontheweb:
@onovotny I think it's pretty ubiquitous now - had to do a lot of testing and support for it when we wrote our OpenTracing adapter for App Insights last year: https://github.com/petabridge/Petabridge.Tracing.ApplicationInsights

onovotny:
Here's what I see if I create one in the portal:

image
image
image

Maybe those should be in a restriction in the template? My aim is to prevent someone from failing due to AI not being in the same region as everything else... or at least make them pick a valid option.

PR #98, opened 2019-08-11
Cleaner temp file handling

benaadams:
Response to twitter 😄

### DIFF
--- a/src/SignService/Utils/DirectoryUtility.cs
+++ b/src/SignService/Utils/DirectoryUtility.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.ApplicationInsights;
 
 namespace SignService.Utils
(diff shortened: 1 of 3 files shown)


src/SignService/Utils/DirectoryUtility.cs
```diff
@@ -31,6 +38,23 @@ static void DeleteDirectory(string fullPath)
             }
         }
 
+        static async Task DeleteDirectoryAsync(string fullPath)
+        {
+            if (!Directory.Exists(fullPath))
```
bruno-garcia: Is this needed? Doesnt delete noop if it doesn't exist?

Tyrrrz: It is needed, you're thinking of CreateDirectory, that one noops

bruno-garcia: You're right [windows](https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/FileSystem.Windows.cs#L166-L177) and [*nix](https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/FileSystem.Unix.cs#L208-L211) throws

PR #200, opened 2020-03-14
Fix build breaks

### LINKED ISSUE #199: Build break on Registry types
I don't know how this is building in Azure Pipelines. The SignService is targeting netcoreapp3.1 but not referencing the `Microsoft.Win32.Registry` package. So local builds fail with this:

```
Utils\mansign2.cs(877,27): error CS0103: The name 'Registry' does not exist in the current context [D:\git\SignService\src\SignService\SignService.csproj] Utils\mansign2.cs(881,33): error CS0103: The name 'RegistryValueKind' does not exist in the current context [D:\git\SignService\src\SignService\SignService.csproj]
Utils\mansign2.cs(881,68): error CS0103: The name 'RegistryValueKind' does not exist in the current context [D:\git\SignService\src\SignService\SignService.csproj]
```

Is this really a configuration fail on my part somehow?

I'd like to send a PR that fixes the build break on my machine. I'd just feel better knowing how it works for anybody else.

AArnott:
This removes the requirement of installing the net47 targeting pack, and adds the missing package reference for accessing the registry in .NET Core.

Fixes #199

### DIFF
--- a/src/Directory.build.props
+++ b/src/Directory.build.props
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
     <PackageReference Include="Nerdbank.GitVersioning" Version="3.1.68" PrivateAssets="all" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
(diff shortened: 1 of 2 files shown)


clairernovotny:
Thanks for this. Is the 47 packageref needed as what we need are the tools from the net 4.8 directory: https://github.com/dotnet/SignService/blob/master/src/SignService/SignService.csproj#L20

AArnott:
No. The ReferenceAssemblies package ref is needed for this:

https://github.com/dotnet/SignService/blob/c5c313881747d2e44b770dc3546c518e0eba0ecd/src/InstallUtility/InstallUtility.csproj#L5

clairernovotny:
Ah, right forgot. Thanks.

PR #201, opened 2020-03-14
Fix compiler warnings

AArnott:
This resolves all warnings in the repo, and changes the pipeline to block introduction of new warnings in the future so we stay warning clean.


clairernovotny:
Thanks. I believe the dotnet pipelines task *just* started properly capturing warnings.

PR #346, opened 2021-02-23
Update NuGet.Config

Youssef1313:


clairernovotny:
not sure why you're deleting the private feed as we consume packages from there that aren't on nuget.org.

Youssef1313:
@clairernovotny See [dependency confusion attack](https://aka.ms/pkg-sec-wp).
If some packages are not on nuget.org, you need to mirror them in the private feed to mitigate the attack.

clairernovotny:
That's not possible with public feeds and Open Source projects can't use private feeds. This is not subject to that attack as the other feed is fully trusted.

Youssef1313:
@clairernovotny If the private feed has packages that don't exist on nuget.org, an attacker could upload his own package with that name on nuget.org.

Could you check please if the build fails as a result of this change, and which package(s) cause it to fail?

PR #405, opened 2021-07-28
Add fallback logic for appx tag

yowl:
This PR adds logic for finding the tag with the `Publisher` attribute for `.appx` files which has a different element name to `.appxbundle` files.  

It does this via a simple `??` fallback if the `appxbundle` tag cannot be found.


yowl:
Fixes an issue in https://github.com/dotnet/SignService/issues/402

PR #433, opened 2022-11-03
Initial port from service to CLI

dtivel:
The CLI is in rough shape:  very little in the way of argument validation and actionable feedback.  You're more likely to get an exception than a friendly error at this point.  I will handle this in a subsequent PR.

I verified that, when passed the right arguments, the CLI does Authenticode sign a binary successfully.  I have not yet verified other scenarios.

This builds and unit tests run successfully.  ~~Note that `dotnet restore bootstrap.proj` is a currently manual build prestep.~~

CC @clairernovotny


clairernovotny (commented):
LGTM so far!

dtivel:
@clairernovotny, fixed console logging not displaying and simplified tools copying.

PR #434, opened 2022-11-04
Add spec for CLI

clairernovotny:
This is the draft spec for the CLI signing tool.


docs/signing-tool-spec.md
```diff
@@ -0,0 +1,128 @@
...
+- **Security:** Code signing is a sensitive operation that requires credentials/secrets. We do not want secrets to be in the MSBuild pipeline as that makes any logs contain those secrets. Ideally, a CI pipeline should contain a separate stage for code signing to ensure that credentials are never unintentionally exposed to a build stage.
+- **Platform:** Authenticode is currently limted to Windows. Thus, while it's possible to sign a NuGet or VSIX cross-platform, the DLL's inside can't be signed unless running on Windows. With the NuGet packages being developer-only artifacts--they're not shipped with the apps--it's critical that the DLL's inside are also signed. Builds for binaries may run on any platform, but as signing is a discreet step in most CI pipelines, it's possible to use a Windows agent for this task.
+
+### Roadmap
+
+There is an impending deadline in December 2022 when the Azure Active Directory Graph v1 endpoints sunset. The scope of the preview release will be limited to the existing funtionality currently in the service. The remaining functionality will be delivered for the release.
```
richlander, line 55: It isn't clear why this is relevant. I'd not go into this detail. I'd just say your actual dates and then have a note that the .NET Foundation has a dependency on this tool being delivered by EOY. Try to avoid crossing streams that don't need to be crossed.

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,128 @@
...
+**Goals**
+
+- Support for Authenticode, VSIX, NuGet, ClickOnce
+- Only run on Windows x64.
+- Support a single certificate for all files in the operation.
+
+**Non-Goals**
+
+- Strong Name signing won't be in v1; guidance is to use an snk not based on a cert. If easy, perhaps can revisit.
+- MicroBuild/ESRP out of scope for v1. Primary audience is end users that have a certificate and "basic" signing needs.
```
richlander, line 68: Too specific. How about "Support for Microsoft internal builds and signing services."

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,128 @@
...
+- Containers of any kind
```
richlander, line 69:
```suggestion
- Containers, including Notary v2 support.
```

docs/signing-tool-spec.md
richlander, line 68:
```suggestion
- Support for Microsoft internal builds and signing services.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+While the third party tools can help, they leave the complicated work of signing the files in the right order to each user, and support for cloud HSMs is limited to Key Vault. With the [announcement](https://techcommunity.microsoft.com/t5/security-compliance-and-identity/azure-code-signing-democratizing-trust-for-developers-and/ba-p/3604669) of Azure Code Signing, and the move towards HSMs, there's a need to support multiple code signing providers in our signing tools.
+
+## Challenges
+
+There are a few challenges around code signing:
+
+### Local Certificates
+
+Today the code signing tooling in the Windows and .NET SDK uses PFX (public/private certificate key pair files or the local certificate store for obtaining certificates. There risks to this approach:
```
joelverhagen, line 27:
```suggestion
Today the code signing tooling in the Windows and .NET SDK uses PFX (public/private certificate key pair files or the local certificate store for obtaining certificates). There risks to this approach:
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+- **Performance:** During a build, many binary artifacts are created that need to be signed. A multi-targeted NuGet package may contain several `.dll` files. An application will likely contain more than one file that needs to be signed. It's much more efficient to pass them all to a signing tool where parallelism is possible than to sign during the inner-loop.
+- **Security:** Code signing is a sensitive operation that requires credentials/secrets. We do not want secrets to be in the MSBuild pipeline as that makes any logs contain those secrets. Ideally, a CI pipeline should contain a separate stage for code signing to ensure that credentials are never unintentionally exposed to a build stage.
+- **Platform:** Authenticode is currently limted to Windows. Thus, while it's possible to sign a NuGet or VSIX cross-platform, the DLL's inside can't be signed unless running on Windows. With the NuGet packages being developer-only artifacts--they're not shipped with the apps--it's critical that the DLL's inside are also signed. Builds for binaries may run on any platform, but as signing is a discreet step in most CI pipelines, it's possible to use a Windows agent for this task.
```
joelverhagen, line 51:
```suggestion
- **Platform:** Authenticode is currently limited to Windows. Thus, while it's possible to sign a NuGet or VSIX cross-platform, the DLL's inside can't be signed unless running on Windows. With the NuGet packages being developer-only artifacts--they're not shipped with the apps--it's critical that the DLL's inside are also signed. Builds for binaries may run on any platform, but as signing is a discreet step in most CI pipelines, it's possible to use a Windows agent for this task.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+- `-u | --descriptionUrl` - Description Url
+- `-td | --timestamp-digest` - Used with the -tr switch to request a digest algorithm used by the RFC 3161 timestamp server.
+- `-kvu | --azure-key-vault-url` - The URL to an Azure Key Vault.
+- `-kvt | --azure-key-vault-tenant-id` - The Tenant Id to authenticate to the Azure Key Vault..
+- `-kvi | --azure-key-vault-client-id` - The Client ID to authenticate to the Azure Key Vault.
+- `-kvs | --azure-key-vault-client-secret` - The Client Secret to authenticate to the Azure Key Vault.
+- `-kvc | --azure-key-vault-certificate` - The name of the certificate in Azure Key Vault.
+- `-kvm | --azure-key-vault-managed-identity` - Use a Managed Identity to access Azure Key Vault.
+
+Signing an archive type (`.zip`, `.nupkg`, `.vsix`) will open up the archive and sign any supported file types. It is strongly recommended to use the `filter` parameter to explicitly list the files inside the archive that should be signed. Signing is recursive; it will sign contents of any detectected `Zip`, `NuPkg` or `VSIX` files inside the uploaded one. After signing contents of the archive, the archive itself is signed if supported (currently `VSIX`).
```
joelverhagen, line 110: I'm curious how the "archive type" is detected. Magic bytes at the beginning of the file? File extension?

joelverhagen, line 110: I don't see a `filter` parameter in the list of parameters.

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+- `-o | --output` - The output file. If omitted, overwrites input. Must be a directory if multiple files are specified.
+- `-b | --base-directory` - Base directory for files to override the working directory.
+- `-f | --force` - Overwrites a signature if it exists.
+- `-m | --max-concurrency` - Maximum concurrency (default is 4)
+- `-fl | --filelist` - Path to file containing paths of files to sign within an archive
+- `-fd | --file-digest` - The digest algorithm to hash the file with.
+- `-tr | --timestamp-rfc3161` - Specifies the RFC 3161 timestamp server's URL.
+- `-d | --description` - Description
+- `-u | --descriptionUrl` - Description Url
```
joelverhagen, line 101: It seems like details of specific signing types will leak into the option list. For example `description` and `descriptionUrl` seem to be related to ClickOnce and would be unused on a .nupkg. What does the tool do if you provide a description but you're just signing a .nupkg? Will it error out? Ignore the parameter? How is it clear to the person running the command that they shouldn't pass that parameter if they're just signing a .nupkg?

I think more generally, I wonder how this parameter organization scales. It seems like there should be a clear way of denoting parameters that are specific to a certain signing type vs. parameters that apply universally. Example: the `-kv*` parameters seem like general ones but the `--description*` ones seems specific to ClickOnce.

Maybe prefix the parameter with `--click-once-description` and `--click-once-description-url`?

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+- Support for Authenticode, VSIX, NuGet, ClickOnce
+- Only run on Windows x64.
+- Support a single certificate for all files in the operation.
+
+**Non-Goals**
+
+- Strong Name signing won't be in v1; guidance is to use an snk not based on a cert. If easy, perhaps can revisit.
+- Support for Microsoft internal builds and signing services.
+- Containers, including Notary v2 support.
+- Extensibility. v1 will support different signing providers.
```
joelverhagen, line 70: It seems like there are (at least) two extensibility pivots: one is the signing provider (the thing that holds the certificate/HSM/secret) and one is the artifact type to be signed. I see the former is planned for extensibility, what about the latter? How would someone extend this tool to sign their own thing (e.g. other package manager artifacts)?

docs/signing-tool-spec.md
clairernovotny: It's a list of specific types--you can see them here: https://github.com/dotnet/sign/blob/cli/src/Sign.Core/Containers/ContainerProvider.cs#L36-L60

joelverhagen: I see .snupkg on there, NuGet.org does not verify .snupkg signatures today and instead depends on the .snupkg PDB matching the .nupkg DLLs. The consumer's debugger does not fetch the .snupkg signature so the benefit is not clear. So I think having this tool sign .snupkg files is a bit strange and is likely unnecessary.

docs/signing-tool-spec.md
clairernovotny: The latter isn't currently planned for external extensibility. Given the structure of the code, potential nesting of containers and file types, an internal integration might be the best option. In that case, a pull request with the to add the artifact type would be the best option. The codebase already has an interface that it'd implement and a way to declare file extensions it supports.

docs/signing-tool-spec.md
clairernovotny: The description and description url are used by Authenticode and ClickOnce. I think it's fair to document what common paramters are used by which file types...

That said, they should be used when signing a NuPkg because in that case, they're not just signing the Nupkg but also the DLL's in side, and Authenticode uses those values.

Those common parameters are tied to the code SubCommand whereas the provider-specific ones (all the azure key vault ones) are declared by the AzureKeyVault subcommand. In 1.0, there'd be three providers, `CertificateStore` (in the binary), and externally via the extensibility `AzureKeyVault`, and `AzureCodeSigning`. Each would declare the parameters they need. Additional providers for additional backends would do the same.

docs/signing-tool-spec.md
clairernovotny: fixed

clairernovotny: It was included for completeness as snupkg is just another NuGet package.

docs/signing-tool-spec.md
joelverhagen: No all nupkg files have DLLs (e.g. meta packages, content packages). I guess what I'm getting at is that this set of parameters is the union of the signing inputs for ALL artifact types. As the number of artifact types grows, this might become an unwieldy or confusing list. Prefixing the parameter name might be a way to make things a bit more organized or clear how values are being used.

Consider if/when repo signing is added. There are "owners" and "service index" URL metadata values that are embedded in the repository signature. If the parameter is just `--owners` or `--service-index` this may be confusing if you don't plan on really needing NuGet repo signing. However `--nupkg-repo-owners` or `--nupkg-repo-service-index` is a bit clearer.

docs/signing-tool-spec.md
joelverhagen: Makes sense. It could be worth calling that out in the spec.

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Roadmap
+
+The scope of the preview release will be limited to the existing funtionality currently in the service. The remaining functionality in this spec will be delivered for the initial release. The .NET Foundation has a dependency on this tool being delivered by the end of 2022.
+
+#### Preview
+
+**Goals**
+
+- Support for Authenticode, VSIX, NuGet, ClickOnce
```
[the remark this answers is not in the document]
joelverhagen: What would the parameters for that look like? Essentially you can have two different signing operations for a single artifact type. How would you (as the caller of the CLI) say "just author sign" or "just repo sign" or "author and repo sign" (that last one probably need not be supported).

This problem may translate to strong-name and authenticode signing (I saw strong-name mentioned as something that may happen later so I wondered how it would be faciliated). Both are applicable to .NET DLLs. How do callers say "do both of these signing operations on a DLL"?

docs/signing-tool-spec.md
joelverhagen: Okay, should it be called out in the spec?

Having the tool signing something that no one verifies could be confusing to the user.

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+There are many HSM cloud services, including Azure Key Vault, that meet the updated key storage requirements, however we do not have first-party support for signing code with those services. There are third-party community solutions to fill this gap: [AzureSignTool](https://github.com/vcsjones/AzureSignTool), [NuGetKeyVaultSignTool](https://github.com/novotnyllc/NuGetKeyVaultSignTool), [OpenOpcSignTool](https://github.com/vcsjones/OpenOpcSignTool), and the .NET Foundation Signing Service (which uses these plus adds additional supported file formats and orchestrates signing the the various types in the right order).
```
skofman1, line 17: Since this is a public facing doc, first-party / third-party should be replaced with Microsoft / community / industry / etc.

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Orchestration
+
+An application/library package typically contains multiple assets that need to be signed. For example, a NuGet package (`.nupkg` file) contains `.dll` files that also must be signed. An ClickOnce or MSIX package also contains `.dll` or `.exe` that need to be signed. A `.vsix` file can contain `.dll` files, and `.nupkg` file that must be signed. These files need to be signed "inside out" to ensure the proper sequence. That is, a `.vsix` containing a `.nupkg` needs to extract the inner `.nupkg` to sign the contained `.dll` files, then sign the `.nupkg` and any other `.dll` files, then repack and sign the `.vsix`. Other types, like ClickOnce and MSIX may contain manifest files that also must be updated during these operations.
+
+With several different code signing tools required, each with its own command line parameters and defaults. The process of code signing is error-prone and hard to get right. The signing tool addresses these challanges by unifying the interface into a single set of options.
```
skofman1, line 41: 'The signing tool addresses these challenges by unifying the interface into a single set of options.'

Since this is a proposal for such a tool, consider rephrasing to something like: 'a signing tool that orchestrates signing of different file types using a single command, will simplify this process significantly, reducing engineering investments by individuals and teams, and expand the usage of code signing.'

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Plan
```
skofman1, line 43: Proposal

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+**Goals**
+
+- Support for Authenticode, VSIX, NuGet (author signature), ClickOnce
+- Only run on Windows x64.
+- Support a single certificate for all files in the operation.
+
+**Non-Goals**
+
+- Strong Name signing won't be in v1; guidance is to use an snk not based on a cert. If easy, perhaps can revisit.
+- Support for Microsoft internal builds and signing services.
```
skofman1, line 68: Since the audience for this spec & tool is external, we shouldn't mention Microsoft internal builds and such (alternatively, rephrase)

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+With several different code signing tools required, each with its own command line parameters and defaults. The process of code signing is error-prone and hard to get right. The signing tool addresses these challanges by unifying the interface into a single set of options.
+
+## Plan
+
+Create a modern signing tool to eventually replace the existing tools. The tool will handle all of our first party signing formats, orchestrate signing files in the right order, and have extensibility to support multiple signing providers. As our customers use a variety of clouds and HSMs, the extensibility will enable us to meet customers needs wherever they store their certificates.
```
skofman1, line 45: While having this tool replace existing tools is a good north star, I think we should hold-off on mentioning it until the tool gains ground with customers.

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+- **Security:** Code signing is a sensitive operation that requires credentials/secrets. We do not want secrets to be in the MSBuild pipeline as that makes any logs contain those secrets. Ideally, a CI pipeline should contain a separate stage for code signing to ensure that credentials are never unintentionally exposed to a build stage.
+- **Platform:** Authenticode is currently limited to Windows. Thus, while it's possible to sign a NuGet or VSIX cross-platform, the DLL's inside can't be signed unless running on Windows. With the NuGet packages being developer-only artifacts--they're not shipped with the apps--it's critical that the DLL's inside are also signed. Builds for binaries may run on any platform, but as signing is a discreet step in most CI pipelines, it's possible to use a Windows agent for this task.
+
+### Roadmap
+
+The scope of the preview release will be limited to the existing funtionality currently in the service. The remaining functionality in this spec will be delivered for the initial release. The .NET Foundation has a dependency on this tool being delivered by the end of 2022.
```
skofman1, line 55: I think folks reading this will not have any prior knowledge about the existing service. Consider expanding about the existing service in the Background section:
1. What the service does
2. Why customers love the service, how do they use it.

In the plan section, talk about why converting the service to a tool is a good idea. Will it make using it easier? Will it solve customer problems with the service? What else?

skofman1, line 55: > The .NET Foundation has a dependency on this tool being delivered by the end of 2022.

We can update that to June 2023

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+- PE files & certain scripts via Authenticode (dll, exe, ps1, sys)
+- MSIX via Authenticode (msix, msixbundle) & related manifests
+- Visual Studio Extensions (VSIX) via Open Packaging Convention
+- ClickOnce & VSTO via Mage (XML Digital Signatures)
+- NuGet Packages
+
+Today each of these areas has their own tools (SignTool, VISXSignTool, Mage, NuGet) to create the signatures. Each tool has its own set of parameters and are written to assume use of the local Certificate Store API's by default. Without any common implementation, any changes to the code signing landscape requires updates to all of the tools. In May 2022, the CA/B forum updates its baseline requirements to require that all new code signing certificates issued after June 2023 use hardware security modules (HSMs) to prevent key theft. While some HSMs contain CSP/KSP support to expose certificates through the CertStore API's, they frequently contain significant limitations, such as requiring an interactive session to authenticate to the device. This makes signing code in the cloud, on build agents, extremely difficult for mainline scenarios.
```
dtivel, line 15:
```suggestion
Today each of these areas has their own tools (SignTool, VISXSignTool, Mage, NuGet) to create signatures. Each tool has its own set of parameters and are written to assume use of the local certificate store API's by default. Without a shared implementation, a new code signing requirement can require individual updates to each tool. In May 2022, the CA/Browser Forum updated its [baseline requirements for publicly trusted code signing certificates](https://cabforum.org/wp-content/uploads/Baseline-Requirements-for-the-Issuance-and-Management-of-Code-Signing.v3.2.pdf) to require that all new code signing certificates issued after June 2023 use hardware security modules (HSM's) to prevent private key theft. While some HSM's contain CSP/KSP support to expose certificates through Windows' certificate store API's, they frequently contain significant limitations, such as requiring an interactive session to authenticate to the device. This makes signing code in the cloud and on build agents extremely difficult for mainline scenarios.
```

docs/signing-tool-spec.md
dtivel, line 17:
```suggestion
There are many HSM cloud services, including Azure Key Vault, that meet the updated key storage requirements, however we do not have first-party support for signing code with those services. There are open source community solutions to fill this gap, such as:

* [AzureSignTool](https://github.com/vcsjones/AzureSignTool)
* [NuGetKeyVaultSignTool](https://github.com/novotnyllc/NuGetKeyVaultSignTool)
* [OpenOpcSignTool](https://github.com/vcsjones/OpenOpcSignTool)

The [.NET Foundation Signing Service](https://github.com/dotnet/sign/tree/legacy-service/servicing) builds on these solutions, adds additional supported file formats, and orchestrates signing the various file types in the right order.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+There are many HSM cloud services, including Azure Key Vault, that meet the updated key storage requirements, however we do not have first-party support for signing code with those services. There are third-party community solutions to fill this gap: [AzureSignTool](https://github.com/vcsjones/AzureSignTool), [NuGetKeyVaultSignTool](https://github.com/novotnyllc/NuGetKeyVaultSignTool), [OpenOpcSignTool](https://github.com/vcsjones/OpenOpcSignTool), and the .NET Foundation Signing Service (which uses these plus adds additional supported file formats and orchestrates signing the the various types in the right order).
+
+While the third party tools can help, they leave the complicated work of signing the files in the right order to each user, and support for cloud HSMs is limited to Key Vault. With the [announcement](https://techcommunity.microsoft.com/t5/security-compliance-and-identity/azure-code-signing-democratizing-trust-for-developers-and/ba-p/3604669) of Azure Code Signing, and the move towards HSMs, there's a need to support multiple code signing providers in our signing tools.
```
dtivel, line 19:
```suggestion
While existing community solutions help, they leave the complicated work of signing the files in the right order to each user and support only Azure Key Vault. With the [announcement](https://techcommunity.microsoft.com/t5/security-compliance-and-identity/azure-code-signing-democratizing-trust-for-developers-and/ba-p/3604669) of Azure Code Signing and the move towards HSM's, there's a need to support multiple code signing providers in our signing tools.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+There are a few challenges around code signing:
+
+### Local Certificates
+
+Today the code signing tooling in the Windows and .NET SDK uses PFX (public/private certificate key pair files or the local certificate store for obtaining certificates). There risks to this approach:
+
+- PFX files are targets in data breaches; their passwords can be cracked
+- Certificates in a local store can be used by any app/malware
+- There’s no revocation mechanism for the users’ access to the certificate; they always have it
```
dtivel, line 31:
```suggestion
- There’s no revocation mechanism for a user's access to the certificate; they always have it
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+- No auditing of signing operations possible
+- EV code signing certificates aren’t easily supported as they require FIPS 140-2 hardware devices with drivers
+
+In May 2022, the CA/B forum updated its baseline requirements to require HSMs so local certificates will no longer be issued for publically trusted roots. The only support the current signing tools have for this scenario is via CSP/KSP drivers provided by some HSM vendors and those do not work well for cloud-based build agents. The current tools would need new investment to support different backends.
```
dtivel, line 35:
```suggestion
In May 2022, the CA/Browser Forum updated its baseline requirements to require HSM's so local certificates will no longer be issued for publicly trusted code signing certificates. The only support the current signing tools have for this scenario is via CSP/KSP drivers provided by some HSM vendors, and those do not work well for cloud-based build agents. The current tools would need new investment to support different backends.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Orchestration
+
+An application/library package typically contains multiple assets that need to be signed. For example, a NuGet package (`.nupkg` file) contains `.dll` files that also must be signed. An ClickOnce or MSIX package also contains `.dll` or `.exe` that need to be signed. A `.vsix` file can contain `.dll` files, and `.nupkg` file that must be signed. These files need to be signed "inside out" to ensure the proper sequence. That is, a `.vsix` containing a `.nupkg` needs to extract the inner `.nupkg` to sign the contained `.dll` files, then sign the `.nupkg` and any other `.dll` files, then repack and sign the `.vsix`. Other types, like ClickOnce and MSIX may contain manifest files that also must be updated during these operations.
```
dtivel, line 39:
```suggestion
An application/library package typically contains multiple assets that need to be signed. For example, a NuGet package (`.nupkg` file) contains `.dll` files that also must be signed. A ClickOnce or MSIX package also contains `.dll` or `.exe` files that need to be signed. A `.vsix` file can contain `.dll` files and `.nupkg` file that must be signed. These files need to be signed "inside out" to ensure the proper sequence. That is, a `.vsix` containing a `.nupkg` needs to extract the inner `.nupkg` to sign the contained `.dll` files, then sign the `.nupkg` and any other `.dll` files, then repack and sign the `.vsix`. Other types, like ClickOnce and MSIX may contain manifest files that also must be updated during these operations.
```

docs/signing-tool-spec.md
dtivel, line 41:
```suggestion
To properly sign all assets, multiple signing tools must be used, and each tool has its own command line syntax, options, and default. The process of code signing is error-prone and hard to get right. The signing tool addresses these challenges by unifying the interface into a single set of options.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+While some of this could be done via an MSBuild task, a CLI tool is preferable to MSBuild tasks for a couple reasons:
+
+- **Performance:** During a build, many binary artifacts are created that need to be signed. A multi-targeted NuGet package may contain several `.dll` files. An application will likely contain more than one file that needs to be signed. It's much more efficient to pass them all to a signing tool where parallelism is possible than to sign during the inner-loop.
+- **Security:** Code signing is a sensitive operation that requires credentials/secrets. We do not want secrets to be in the MSBuild pipeline as that makes any logs contain those secrets. Ideally, a CI pipeline should contain a separate stage for code signing to ensure that credentials are never unintentionally exposed to a build stage.
```
dtivel, line 50:
```suggestion
- **Security:** Code signing is a sensitive operation that requires credentials/secrets. Use of these secrets should be as limited as possible to prevent leakage into the rest of a build pipeline, such as log files or unrelated build tasks. Ideally, a CI pipeline should contain a separate stage for code signing to ensure that credentials are never unintentionally exposed to a build stage.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+- **Platform:** Authenticode is currently limited to Windows. Thus, while it's possible to sign a NuGet or VSIX cross-platform, the DLL's inside can't be signed unless running on Windows. With the NuGet packages being developer-only artifacts--they're not shipped with the apps--it's critical that the DLL's inside are also signed. Builds for binaries may run on any platform, but as signing is a discreet step in most CI pipelines, it's possible to use a Windows agent for this task.
```
dtivel, line 51:
```suggestion
- **Platform:** Authenticode is currently limited to Windows. Thus, while it's possible to sign a NuGet or VSIX cross-platform, the DLL's inside can't be signed unless running on Windows. With the NuGet packages being developer-only artifacts--they're not shipped with the apps--it's critical that the DLL's inside are also signed. Builds for binaries may run on any platform, but as signing is a discreet step in most CI pipelines, it's reasonable to require a Windows build agent for this task.
```

docs/signing-tool-spec.md
dtivel, line 55:
```suggestion
The scope of the preview release will be limited to the existing funtionality currently in the service. The remaining functionality in this spec will be delivered in a later 1.0 release. The .NET Foundation has a dependency on this tool being delivered by [June 30, 2023](https://learn.microsoft.com/en-us/answers/questions/768833/when-is-adal-and-azure-ad-graph-reaching-end-of-li.html).
```

docs/signing-tool-spec.md
dtivel, line 45:
```suggestion
Create a modern signing tool to eventually replace the existing tools. The tool will handle all of our first party signing formats, orchestrate signing files in the right order, and have extensibility to support multiple raw signature providers. As our customers use a variety of clouds and HSM's, the extensibility will enable us to meet our customers' needs wherever they store their certificates.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+- Only run on Windows x64.
+- Support a single certificate for all files in the operation.
+
+**Non-Goals**
+
+- Strong Name signing won't be in v1; guidance is to use an snk not based on a cert. If easy, perhaps can revisit.
+- Support for Microsoft internal builds and signing services.
+- Containers, including Notary v2 support.
+- Extensibility. v1 will support different signing providers.
+- Support for platforms other than Windows x64. Future work will be required to support ARM64 and non-Windows hosts. Support for certain file types may be limited due to platform support.
```
dtivel, line 71:
```suggestion
- Support Authenticode on platforms other than Windows x64. Future work will be required to support ARM64 and non-Windows hosts. Support for certain file types may be limited due to platform support.
```

docs/signing-tool-spec.md
```diff
@@ -0,0 +1,129 @@
...
+**Goals**
+
+- Extensibility mechanism to support different code signing providers with a dynamic lookup so the core client remains agnostic of the backend
+- Offline distribution for core plus backend provider
+- Three providers: Certificate Store, Azure Key Vault, Azure Code Signing
+- Support for additional formats: [.HLKX](https://github.com/dotnet/sign/issues/422), [VBA](https://github.com/dotnet/sign/issues/364)
+- Verification of signatures
+
+
+### Inputs
```
dtivel, line 85: From here to the end doesn't belong in this spec.  It belongs in a separate CLI syntax and options spec.

dtivel (changes requested):

dtivel (approved):

PR #446, opened 2022-11-12
CLI:  create .NET tool package

dtivel:
Resolve https://github.com/dotnet/sign/issues/445.

This change creates a .NET tool package.

Note:  this source branch is based on a target branch which currently has an open PR (https://github.com/dotnet/sign/pull/444).  The reason is because the cli branch build is currently broken and CI builds for PR's fail.  The current target branch has the fix.  Once the target branch merges into the cli branch, I will retarget this PR at the cli branch.


joelverhagen (approved):
The pack properties look good. The Version.Details.xml file and eng/Version.props are not something I'm familiar with (Arcade maybe?) so I'll leave those to another person to review.

dtivel:
Yes, both files are from [dotnet/Arcade](https://github.com/dotnet/arcade/blob/main/Documentation/Onboarding.md).

When I onboarded to Arcade, I accidentally [copied Version.Details.xml into the root directory](https://github.com/dotnet/sign/commit/250ed266116a9b40960e11e0d1f5bf40c2185b8b#diff-4b772e14fcdd6149440e6ce851a5ac65146fc8914375feed748f8eaaa5be61b8) --- in addition to where it should go, in the eng\ directory --- and checked it in.  This file (at the root) isn't used and should just be deleted.

eng/Version.props is just versioning.  @clairernovotny and I already discussed this versioning change, so it's already had a verbal review.

Issue #447, opened 2022-11-14
CLI:  containers not opened


dtivel:
While writing tests, I found a bug where containers were not opened prior to querying their contents.

PR #448, opened 2022-11-14
CLI:  fix bugs and add tests

dtivel:
Resolve https://github.com/dotnet/sign/issues/447.

This change adds a bunch of tests and fixes a few bugs found in the process.

Note:  this source branch is based on a target branch which currently has an open PR (https://github.com/dotnet/sign/pull/446).  The reason is because the cli branch build is currently broken and CI builds for PR's fail.  The current target branch has the fix.  Once the target branch merges into the cli branch, I will retarget this PR at the cli branch.


test/Sign.Core.Test/TestInfrastructure/AggregatingSignatureProviderTest.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// Creates a test for testing <see cref="AggregatingSignatureProvider" />.
+        /// </summary>
+        /// <param name="paths">Zero or more relative file or directory paths.</param>
+        /// <remarks>
+        /// All paths:
+        ///
+        ///     * must use only forward slashes (/) to delimit files and directories
+        ///
+        ///         Examples:
+        ///           ❌ directory\a.dll
```
joelverhagen, line 26: Emojis in comments 💯

test/Sign.Core.Test/TestInfrastructure/ContainerSpy.cs
```diff
@@ -0,0 +1,58 @@
+using Microsoft.Extensions.FileSystemGlobbing;
+
+namespace Sign.Core.Test
+{
+    internal sealed class ContainerSpy : IContainer
```
joelverhagen, line 5: No 🕵️ emoji? 😄

test/Sign.Core.Test/SignatureProviders/AppInstallerServiceSignatureProviderTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            ArgumentNullException exception = Assert.Throws<ArgumentNullException>(
+                () => new AppInstallerServiceSignatureProvider(
+                    Mock.Of<IKeyVaultService>(),
+                    logger: null!));
+
+            Assert.Equal("logger", exception.ParamName);
+        }
+
+        [Theory]
+        [InlineData(".appinstaller")]
```
joelverhagen, line 41: Is culture testing like turkish i interesting here (case insensitivity has varying behaviors across culture setting)? Or is it not worth the effort?

test/Sign.Core.Test/SignatureProviders/AggregatingSignatureProviderTests.Containers.cs
```diff
@@ -0,0 +1,235 @@
+namespace Sign.Core.Test
```
joelverhagen, line 1: Should this code have the .NET Foundation header?

src/Sign.Core/ServiceProvider.cs
```diff
@@ -20,23 +20,6 @@ public ServiceProvider(IServiceProvider serviceProvider)
             return _serviceProvider.GetService(serviceType);
         }
 
-        internal static ServiceProvider CreateForCli(LogLevel logLevel = LogLevel.Information)
```
joelverhagen, line 23: Unused code, I guess?

joelverhagen (approved):
LGTM

src/Sign.Core/ServiceProvider.cs
dtivel: Yep, unused.

test/Sign.Core.Test/SignatureProviders/AppInstallerServiceSignatureProviderTests.cs
clairernovotny: What code do we have that's culture-specific? Our inputs/outputs should all be invariant as they're filenames? If anything, we might want to add the analyzer that checks for specifying the CultureInfo and ensure it's always Invariant?

test/Sign.Core.Test/SignatureProviders/DistinguishedNameParserTests.cs
```diff
@@ -0,0 +1,41 @@
+namespace Sign.Core.Test
+{
+    public class DistinguishedNameParserTests
+    {
+        [Fact]
+        public void Parse_WhenSubjectIsValid_ReturnsRelativeDistinguishedNames()
+        {
+            string subject = "CN=Microsoft Code Signing PCA 2011, O=Microsoft Corporation, L=Redmond, S=Washington, C=US";
```
clairernovotny, line 8: We should have a test for a subject name that contains internal (escaped) quotes. Many parsers fail at that.

test/Sign.Core.Test/SignatureProviders/AppInstallerServiceSignatureProviderTests.cs
joelverhagen: I don't know the full answer to your question :) I asked because I saw a pattern I'm familiar with. Case insensitivity works differently on different cultures. Example: https://dotnetfiddle.net/a3vQK1

I dug into the code and we use the invariant so it should not be culture dependent. 
https://github.com/dotnet/sign/blob/6d16cdda0c06336041ce283fbca65d3aad1683cc/src/Sign.Core/SignatureProviders/AppInstallerServiceSignatureProvider.cs#LL30C13-L30C111

So the product code looks fine, but (theoretically) the test code could have a test for `.appİnstaller` and `.appınstaller` to ensure that `CanSign` returns false. It's also fine to say "that test is too crazy to write" 😄.

The analyzer idea is great. I think there are some global readiness analyzers that NuGet Client team has used recently.

test/Sign.Core.Test/SignatureProviders/AggregatingSignatureProviderTests.Containers.cs
dtivel: See https://github.com/dotnet/sign/issues/449.  I will address this in a separate PR.

test/Sign.Core.Test/SignatureProviders/AppInstallerServiceSignatureProviderTests.cs
dtivel: I pushed a commit that:

1.  Enabled analyzers for several string comparison issues.  Notably, I did not resolve CA1305 warnings or enable CA1305 analyzer warnings; a work item tracking that is https://github.com/dotnet/sign/issues/450.
2.  Changed file name comparison from `InvariantCultureIgnoreCase` to `OrdinalIgnoreCase`, which is consistent with guidance [here](https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings?WT.mc_id=DT-MVP-5003978#choosing-a-stringcomparison-member-for-your-method-call).
3.  Added unit tests, especially for Turkish i.

test/Sign.Core.Test/SignatureProviders/DistinguishedNameParserTests.cs
dtivel: https://github.com/dotnet/sign/issues/451

Issue #469, opened 2022-11-22
CLI:  ClickOnce signature provider broken


dtivel:
While writing tests for ClickOnceSignatureProvider, I found 3 bugs:

1. The `IContainer` isn't opened before trying to query content files.
2. The `RSAPKCS1SHA256SignatureDescription` type isn't publicly creatable, which is a `CryptoConfig` requirement.
3. While porting code, I converted file representations from `string` to `FileInfo`.  In a couple places in `ClickOnceSignatureProvider` this produced syntactically correct but semantically incorrect code:  `string.Replace(...)` became `FileInfo.Replace(...)` and failed at runtime.

Issue #470, opened 2022-11-22
CLI:  ClickOnce signing workaround should be removed


dtivel:
@clairernovotny says [this workaround](https://github.com/dotnet/sign/tree/22a3397dcbfb0bd61c2b4ab9db38e101fa8ec06a#clickonce) should be removed.

clairernovotny:
With the tool operating locally, it should accept a directory as an argument and handle all the required files instead of needing them packaged up.

PR #471, opened 2022-11-22
CLI:  fix bugs and add tests

dtivel:
Resolve https://github.com/dotnet/sign/issues/469.

This change fixes the 3 bugs described in the issue and adds unit tests and, in particular, one unit test which thoroughly verifies `ClickOnceSignatureProvider.SignAsync(...)` behavior.

Note that adding `.clickonce` file extension as a supported ZIP container extension follows current behavior in the existing SignService CLI but should be viewed as a temporary change.  For details see https://github.com/dotnet/sign/issues/470.

### DIFF
--- a/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
+++ b/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
@@ -80,53 +80,42 @@ namespace Sign.Core
                     using (TemporaryDirectory temporaryDirectory = new(_directoryService))
                     using (IContainer zip = _containerProvider.GetContainer(file)!)
                     {
+                        await zip.OpenAsync();
+
                         // Look for the data files first - these are .deploy files
                         // we need to rename them, sign, then restore the name
 
-                        IEnumerable<FileInfo> filteredFiles = GetFiles(zip, options).ToList();
-
+                        List<FileInfo> filteredFiles = GetFiles(zip, options).ToList();
                         List<FileInfo> deployFilesToSign = filteredFiles
                             .Where(f => ".deploy".Equals(f.Extension, StringComparison.OrdinalIgnoreCase))
                             .ToList();
                         List<FileInfo> contentFiles = new();
 
-                        foreach (FileInfo dfile in deployFilesToSign)
-                        {
-                            // Rename to file without extension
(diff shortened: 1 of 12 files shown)


src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
```diff
@@ -80,68 +80,57 @@ public async Task SignAsync(IEnumerable<FileInfo> files, SignOptions options)
...
-                        }
+                        RemoveDeployExtension(deployFiles, contentFiles);
 
                         // at this point contentFiles has all deploy files renamed
 
                         // Inner files are now signed
                         // now look for the manifest file and sign that
 
-                        var manifestFile = filteredFiles.Single(f => ".manifest".Equals(f.Extension, StringComparison.OrdinalIgnoreCase));
+                        FileInfo manifestFile = filteredFiles.Single(f => ".manifest".Equals(f.Extension, StringComparison.OrdinalIgnoreCase));
```
joelverhagen, line 116: nit: consider an introducing some idiom in the codebase that allows succinct file extension checking. Maybe it's a new LINQ method (`SingleExtension` or `WhereExtension`) or a helper method called in existing `LINQ` methods. Or a wrapper around the extension LINQ expressions. Seems a bit verbose as-is. You need to think about the StringComparison each time but really it should be an impl. detail of a helper (IMO. Just a thought!

src/Sign.Core/SignatureProviders/RSAPKCS1SHA256SignatureDescription.cs
```diff
@@ -2,9 +2,11 @@
 
 namespace Sign.Core
 {
-    internal sealed class RSAPKCS1SHA256SignatureDescription : RSAPKCS1SignatureDescription
+    // This type and its default constructor are public because:
+    //   "Algorithms added to CryptoConfig must be accessible from outside their assembly."
```
joelverhagen, line 6: Seems like a requirement of this type? https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.cryptoconfig.addalgorithm?view=net-7.0 (based on exception message). A URL reference could be nice.

joelverhagen (approved):

Issue #472, opened 2022-11-22
CLI:  remove use of SHA-1


dtivel:
Static analysis found 4 cases where SHA-1 is used:

* https://github.com/dotnet/sign/blob/22a3397dcbfb0bd61c2b4ab9db38e101fa8ec06a/src/Sign.Core/Native/mansign.cs#L416
* https://github.com/dotnet/sign/blob/22a3397dcbfb0bd61c2b4ab9db38e101fa8ec06a/src/Sign.Core/Native/mansign.cs#L445
* https://github.com/dotnet/sign/blob/22a3397dcbfb0bd61c2b4ab9db38e101fa8ec06a/src/Sign.Core/Native/mansign2.cs#L358
* https://github.com/dotnet/sign/blob/22a3397dcbfb0bd61c2b4ab9db38e101fa8ec06a/src/Sign.Core/Native/mansign2.cs#L402

PR #473, opened 2022-11-22
CLI:  remove use of SHA-1

dtivel:
Resolve https://github.com/dotnet/sign/issues/472.

This change removes use of SHA-1.  I don't think any of this code was ever called, but SDL prohibits its existence anyway.

Note:  this source branch is based on a target branch which currently has an open PR (https://github.com/dotnet/sign/pull/471).  The reason is that these PR's are intended to be merged in dependency order and chained branching simplifies PR diffs.  Once the target branch merges into the cli branch, I will retarget this PR at the cli branch.


joelverhagen (approved):

src/Sign.Core/Native/mansign2.cs
```diff
@@ -386,28 +364,14 @@ static byte[] ComputeHashFromManifest(XmlDocument manifestDom, bool oldFormat, b
...
                 {
-                    using var sha2 = SHA256.Create();
-                    var hash = sha2.ComputeHash(exc.GetOutput() as MemoryStream);
-                    if (hash == null)
-                    {
-                        throw new CryptographicException(Win32.TRUST_E_BAD_DIGEST);
-                    }
-
-                    return hash;
+                    throw new CryptographicException(Win32.TRUST_E_BAD_DIGEST);
```
kartheekp-ms, line 371: Can we extract this logic into a method so that it can be called here and line no:367 to 371?

dtivel: Good feedback, but I would like to hold off on that change.

This file and mansign.cs were once copies of .NET Framework source code.  See https://github.com/microsoft/referencesource/blob/master/inc/mansign.cs and https://github.com/microsoft/referencesource/blob/master/inc/mansign2.cs.

I've tried to maintain a high bar for making changes to these files.  I've added copyright headers, suppressed some warnings (that would be errors with treat warnings as errors enabled), and deleted code.  I've not generally refactored or improved the code, because long-term I think we should get rid of it and migrate to an actively maintained library that does manifest signing.  See https://github.com/dotnet/sign/issues/479.

dtivel: If you feel strongly about this, let's talk.

PR #476, opened 2022-11-22
CLI:  block x86 with useful error message

dtivel:
Resolve https://github.com/dotnet/sign/issues/475.

The upcoming preview explicitly only supports Windows x64.  This change blocks x86 with a link to upvote support for it.

### DIFF
--- a/src/Sign.Cli/Program.cs
+++ b/src/Sign.Cli/Program.cs
@@ -13,6 +13,13 @@ namespace Sign.Cli
     {
         internal static async Task<int> Main(string[] args)
         {
+            if (!Environment.Is64BitProcess)
+            {
+                Console.Error.WriteLine("Only Windows x64 is supported at this time. See https://github.com/dotnet/sign/issues/474 regarding Windows x86 support.");
+
+                return ExitCode.Failed;
+            }
+
             string directory = Path.GetDirectoryName(Environment.ProcessPath!)!;
             string baseDirectory = Path.Combine(directory, "tools", "SDK", "x64");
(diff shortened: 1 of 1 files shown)


src/Sign.Cli/Program.cs
```diff
@@ -23,6 +23,13 @@ internal static async Task<int> Main(string[] args)
             Kernel32.LoadLibraryW($@"{baseDirectory}\wintrust.dll");
             Kernel32.LoadLibraryW($@"{baseDirectory}\mssign32.dll");
 
+            if (!Environment.Is64BitProcess)
+            {
+                Console.Error.WriteLine("Only Windows x64 is supported at this time.  See https://github.com/dotnet/sign/issues/474 regarding Windows x86 support.");
+
+                return ExitCode.Failed;
+            }
+
```
kartheekp-ms, line 32: should we add this logic to the beginning of this method?

kartheekp-ms (approved):

src/Sign.Cli/Program.cs
dtivel: My reluctance in moving this to the beginning of this method is explained [here](https://github.com/dotnet/sign/pull/476/files#diff-4e7543fbcd4b76878fbcb858fe3748c5df8151894f795bb1db6eafa2381ee391L16-L20).

Presently, I don't have a way of testing x86 Windows to see if this matters.  Azure DevOps only supports bring-your-own-x86-OS; the marketplace doesn't have x86 images available.

Issue #449, opened 2022-11-15
CLI:  add license headers and header enforcement


dtivel:
Add license headers to every file.

Also see https://learn.microsoft.com/en-us/visualstudio/ide/reference/add-file-header?view=vs-2022

PR #477, opened 2022-11-22
CLI:  add license headers and enable analyzer

dtivel:
Resolve https://github.com/dotnet/sign/issues/449.

This change adds license headers to all files and enables an analyzer to check for missing/incorrect headers.

Note:  this source branch is based on a target branch which currently has an open PR (https://github.com/dotnet/sign/pull/471).  The reason is that these PR's are intended to be merged in dependency order and chained branching simplifies PR diffs.  Once the target branch merges into the cli branch, I will retarget this PR at the cli branch.


joelverhagen (approved):
LGTM, just wanted to be sure the notice for non-DNF code is done correctly based on the whatever latest guidance CELA/OSS at Microsoft may have.

PR #486, opened 2022-11-25
CLI:  add 3rd party notices file to package

dtivel:
Resolve https://github.com/dotnet/sign/issues/483.

This change adds https://github.com/dotnet/sign/blob/cli/THIRD-PARTY-NOTICES.txt to the root of the package.

### DIFF
--- a/src/Sign.Cli/Sign.Cli.csproj
+++ b/src/Sign.Cli/Sign.Cli.csproj
@@ -67,6 +67,10 @@
       <Pack>true</Pack>
       <PackagePath>tools\$(TargetFramework)\any\tools</PackagePath>
     </Content>
+    <Content Include="$(RepositoryRootDirectory)\THIRD-PARTY-NOTICES.txt">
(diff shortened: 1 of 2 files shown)


src/Sign.Cli/Sign.Cli.csproj
```diff
@@ -67,6 +67,10 @@
       <Pack>true</Pack>
       <PackagePath>tools\$(TargetFramework)\any\tools</PackagePath>
     </Content>
+    <Content Include="$(RepositoryRootDirectory)\THIRD-PARTY-NOTICES.txt">
```
erdembayar, line 70: Does this file exist yet?

dtivel: Yes, in the [cli](https://github.com/dotnet/sign/tree/cli) branch, which is not the default branch.

erdembayar (approved):

Issue #502, opened 2022-11-29
CLI: warn if certificate is not time valid


dtivel:
@clairernovotny felt that the tool should not hard block on these certificates and should instead just warn.

PR #503, opened 2022-11-29
CLI: warn if certificate is not time valid

dtivel:
Resolve https://github.com/dotnet/sign/issues/502.

If a certificate is not yet valid or is expired, a warning will display "The certificate is not time valid."

### DIFF
--- a/src/Sign.Core/Certificates/CertificateVerifier.cs
+++ b/src/Sign.Core/Certificates/CertificateVerifier.cs
@@ -1,0 +1,34 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Extensions.Logging;
+
+namespace Sign.Core
+{
+    internal sealed class CertificateVerifier : ICertificateVerifier
+    {
+        private readonly ILogger _logger;
+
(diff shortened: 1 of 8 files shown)


src/Sign.Core/Certificates/CertificateVerifier.cs
```diff
@@ -0,0 +1,34 @@
...
+            _logger = logger;
+        }
+
+        public void Verify(X509Certificate2 certificate)
+        {
+            ArgumentNullException.ThrowIfNull(certificate, nameof(certificate));
+
+            DateTime now = DateTime.Now;
+
+            if (now < certificate.NotBefore || certificate.NotAfter < now)
```
erdembayar, line 28: Should we separate them and give different warning for each scenario.

erdembayar (approved):

src/Sign.Core/Certificates/CertificateVerifier.cs
```diff
@@ -0,0 +1,34 @@
...
+            {
+                _logger.LogWarning("The certificate is not time valid.");
```
skofman1: Consider printing the certificate.NotBefore, certificate.NotAfter in the Warning message, to make it clearer for the user

erdembayar: Yes, he did it [here](https://github.com/dotnet/sign/pull/503/commits/423cf60a46f54d771ace32290eeb463138f4bfbd).

PR #522, opened 2022-12-09
Update Arcade

dtivel:
Resolve https://github.com/dotnet/sign/issues/521.

This change does a few things:

* copies from dotnet/arcade the latest version of Arcade
* undoes [this change](https://github.com/dotnet/sign/pull/518) because Arcade included the fix by default ([here](https://github.com/dotnet/sign/blob/f4857cfd8b935c1a2b9f9e837617f3763cafd03b/eng/common/sdk-task.ps1#L67) and [here](https://github.com/dotnet/sign/blob/f4857cfd8b935c1a2b9f9e837617f3763cafd03b/eng/common/tools.ps1#L369))


zivkan:
dnceng's tools have automatic dependency flow, including support for arcade sdk, so that it's automatic and avoid human mistakes: https://github.com/dotnet/arcade/blob/main/Documentation/DependencyFlowOnboarding.md

Should we do that instead of these manual updates?

dtivel:
Great idea.  Thanks for the pointer.  I created https://github.com/dotnet/sign/issues/523 for that.  For now, I think this PR is still good for merging as-is.  It also undoes a previous change which is no longer necessary.

erdembayar (approved):

Issue #538, opened 2022-12-19
Sign fails with missing timestamp


clairernovotny:
This command fails:

`sign code azure-key-vault -d "scratch foo" -u https://the.site -kvi **** -kvs **** -kvt **** -kvu **** -kvc **** .\scratch.1.0.0.nupkg`

```
info: Sign.Core.Signer[0]
      Submitting 'C:\Users\cnov\dev\scratch\bin\Debug\scratch.1.0.0.nupkg' for signing.
info: Sign.Core.Signer[0]
      SignAsync called for C:\Users\cnov\dev\scratch\bin\Debug\scratch.1.0.0.nupkg. Using C:\Users\cnov\AppData\Local\Temp\frxvsoeu.eht\la2t4zkk.nupkg locally.
info: Sign.Core.IDirectoryService[0]
      Extracting container file C:\Users\cnov\AppData\Local\Temp\frxvsoeu.eht\la2t4zkk.nupkg to C:\Users\cnov\AppData\Local\Temp\2ox4xsd1.h5b.
info: Sign.Core.ISignatureProvider[0]
      Signing SignTool job with 1 files
info: Sign.Core.ISignatureProvider[0]
      Signing using C:\Users\cnov\AppData\Local\Temp\2ox4xsd1.h5b\lib\net7.0\scratch.dll
info: Sign.Core.ISignatureProvider[0]
      Sign tool completed successfully
info: Sign.Core.IDirectoryService[0]
      Rebuilding container C:\Users\cnov\AppData\Local\Temp\frxvsoeu.eht\la2t4zkk.nupkg from C:\Users\cnov\AppData\Local\Temp\2ox4xsd1.h5b.
info: Sign.Core.INuGetSignTool[0]
      Signing using C:\Users\cnov\AppData\Local\Temp\frxvsoeu.eht\la2t4zkk.nupkg
fail: Sign.Core.INuGetSignTool[0]
      Value cannot be null. (Parameter 'uriString')
      System.ArgumentNullException: Value cannot be null. (Parameter 'uriString')
         at System.Uri..ctor(String uriString)
[...]

PR #540, opened 2023-01-05
Require timestamping for signing

dtivel:
Resolve https://github.com/dotnet/sign/issues/538.

This change makes timestamping required.

CC @clairernovotny

### DIFF
--- a/src/Sign.Cli/CodeCommand.cs
+++ b/src/Sign.Cli/CodeCommand.cs
@@ -123,7 +123,7 @@ namespace Sign.Cli
                 !(string.Equals(Uri.UriSchemeHttp, uri.Scheme, StringComparison.OrdinalIgnoreCase) ||
                     string.Equals(Uri.UriSchemeHttps, uri.Scheme, StringComparison.OrdinalIgnoreCase)))
             {
-                result.ErrorMessage = $"Unsupported value for --{result.Argument.Name}.  The value must be an HTTP or HTTPS URL.";
+                result.ErrorMessage = $"Unsupported value for --{result.Argument.Name}.  The value must be an absolute HTTP or HTTPS URL.";
 
                 return null;
             }
--- a/src/Sign.Cli/CodeCommand.cs
+++ b/src/Sign.Cli/CodeCommand.cs
@@ -13,14 +13,14 @@ namespace Sign.Cli
     {
         internal Option<DirectoryInfo> BaseDirectoryOption { get; } = new(new[] { "-b", "--base-directory" }, ParseBaseDirectoryOption, description: "Base directory for files to override the working directory.");
         internal Option<string> DescriptionOption { get; } = new(new[] { "-d", "--description" }, "Description of the signing certificate.");
-        internal Option<Uri> DescriptionUrlOption { get; } = new(new[] { "-u", "--description-url" }, "Description URL of the signing certificate.");
+        internal Option<Uri?> DescriptionUrlOption { get; } = new(new[] { "-u", "--description-url" }, ParseUrl, description: "Description URL of the signing certificate.");
         internal Option<HashAlgorithmName> FileDigestOption { get; } = new(new[] { "-fd", "--file-digest" }, ParseHashAlgorithmName, description: "Digest algorithm to hash the file with. Allowed values are sha256, sha384, and sha512.");
         internal Option<string?> FileListOption = new(new[] { "-fl", "--file-list" }, "Path to file containing paths of files to sign within an archive.");
         internal Option<int> MaxConcurrencyOption { get; } = new(new[] { "-m", "--max-concurrency" }, ParseMaxConcurrencyOption, description: "Maximum concurrency (default is 4)");
         internal Option<string?> OutputOption { get; } = new(new[] { "-o", "--output" }, "Output file or directory. If omitted, overwrites input file.");
         internal Option<string?> PublisherNameOption { get; } = new(new[] { "-pn", "--publisher-name" }, "Publisher name (ClickOnce).");
         internal Option<HashAlgorithmName> TimestampDigestOption { get; } = new(new[] { "-td", "--timestamp-digest" }, ParseHashAlgorithmName, description: "Used with the -t switch to request a digest algorithm used by the RFC 3161 timestamp server. Allowed values are sha256, sha384, and sha512.");
-        internal Option<Uri> TimestampUrlOption { get; } = new(new[] { "-t", "--timestamp-url" }, "RFC 3161 timestamp server URL. If this option is not specified, the signed file will not be timestamped.");
+        internal Option<Uri?> TimestampUrlOption { get; } = new(new[] { "-t", "--timestamp-url" }, ParseUrl, description: "RFC 3161 timestamp server URL.");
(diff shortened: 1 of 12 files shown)


src/Sign.Cli/CodeCommand.cs
```diff
@@ -114,5 +115,20 @@ private static HashAlgorithmName ParseHashAlgorithmName(ArgumentResult result)
...
         }
+
+        private static Uri? ParseUrl(ArgumentResult result)
+        {
+            if (result.Tokens.Count != 1 ||
+                !Uri.TryCreate(result.Tokens[0].Value, UriKind.Absolute, out Uri? uri) ||
+                !(string.Equals(Uri.UriSchemeHttp, uri.Scheme, StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(Uri.UriSchemeHttps, uri.Scheme, StringComparison.OrdinalIgnoreCase)))
+            {
+                result.ErrorMessage = $"Unsupported value for --{result.Argument.Name}.  The value must be an HTTP or HTTPS URL.";
```
erdembayar, line 126: Do we need to localize this string? It looks user facing info.

erdembayar: Is there test for this one?  We're mutating the original input to set error and same time returning `null`, not sure if it's working as expected unless there is actual test for this.

test/Sign.Core.Test/SignatureProviders/AggregatingSignatureProviderTests.cs
```diff
@@ -10,7 +10,7 @@ namespace Sign.Core.Test
 {
     public partial class AggregatingSignatureProviderTests
     {
-        private static readonly SignOptions _options = new(HashAlgorithmName.SHA256);
+        private static readonly SignOptions _options = new(HashAlgorithmName.SHA256, new Uri("http://timestamp.test"));
```
erdembayar, line 13: Are there tests for timestamping signing assertion if timestamping server is provided?
If I understand correctly if it's timestamped, then we can verify it against same server, right?

erdembayar (commented):
Test related comments

src/Sign.Cli/CodeCommand.cs
agr, line 126: The error message doesn't seem to match the validations above. There are checks that
* Check number of tokens (instances of an argument, I presume?);
* Check if the URL is valid and is absolute;
* Check if the URL schema is HTTP or HTTPS.
The error message only mentions the last requirement of those 3.

agr: It would be especially confusing to see this message when there is more than one url artument.

test/Sign.Core.Test/SignatureProviders/AggregatingSignatureProviderTests.cs
```diff
@@ -147,7 +147,7 @@ public async Task SignAsync_WhenFilesIsNull_Throws()
             ArgumentNullException exception = await Assert.ThrowsAsync<ArgumentNullException>(
                 () => provider.SignAsync(
                     files: null!,
-                    new SignOptions(HashAlgorithmName.SHA256)));
+                    new SignOptions(HashAlgorithmName.SHA256, new Uri("http://timestamp.test"))));
```
kartheekp-ms, line 150: Can we use the `_options` static variable instead of creating a new `SignOptions` object?

src/Sign.Cli/CodeCommand.cs
dtivel: Great catch!  I missed this string because of the order in which I applied (cherry-picked) a commit.  I fixed it in the other PR which is focused on localizability.  https://github.com/dotnet/sign/pull/543/commits/d72c060ed6a7604fac16e491f0955d308ea3d9eb

dtivel: As far as tests go, yes, there are positive and negative tests [here](https://github.com/dotnet/sign/pull/540/files#diff-1a2ec79601f9b01f7dc6a290224d0c978bc8691aa175c4063d19b1dd0336bd1dR24-R41).  (The tests are shared by all `Uri`-based option tests.)

Setting the `ErrorMessage` property to a non-`null` value generates a parse error and prevents invocation, so a `null` `Uri` will not happen in positive cases.  See System.CommandLine [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.commandline.parsing.symbolresult.errormessage?view=system-commandline#remarks).

The comment [here](https://github.com/dotnet/sign/pull/540/files#diff-bc5f203d5b2bd02ed1b817ea63cc6ab928a45e92dd1b0a259f948c2beb0c1f2dR65-R68), where the value is actually used, also covers this.

test/Sign.Core.Test/SignatureProviders/AggregatingSignatureProviderTests.cs
dtivel: I'm not sure I understand the question, so I'll ask/answer a few questions I think you might be asking.

Q:  Are there automated tests that verify that a (syntactically) valid value was passed for the `--timestamp-url` option?
A:  Yes, [here](https://github.com/dotnet/sign/pull/540/files#diff-1a2ec79601f9b01f7dc6a290224d0c978bc8691aa175c4063d19b1dd0336bd1dR24-R32).

Q:  Are there automated tests that timestamping actually works end-to-end?
A:  No, that level of end-to-end testing isn't automated.

Q:  Does verifying a timestamp access the timestamping server?
A:  Usually, no.  A timestamping server provides a timestamp, and the timestamp contains the signing certificate.  Verifying the timestamp may involve chain building the certificate, to include revocation checking, but neither of those operations typically hit the timestamp server.

If you meant something else, let me know.

src/Sign.Cli/CodeCommand.cs
dtivel: I added "absolute" to the error message (your # 2).

> It would be especially confusing to see this message when there is more than one url artument.

You won't.  System.CommandLine will report `Option '--timestamp-url' expects a single argument but 2 were provided.`

test/Sign.Cli.Test/Options/UriOptionTests.cs
```diff
@@ -20,5 +20,32 @@ public void Option_WhenValueFailsToParse_HasError()
...
+            Verify($"{ShortOption} {expectedValue.OriginalString}", expectedValue);
+        }
+
+        [Theory]
+        [InlineData("//domain.test")]
+        [InlineData("/path")]
+        [InlineData("file:///file.bin")]
+        public void Option_WithShortOptionAndInvalidUrl_HasErrors(string invalidUrl)
+        {
+            VerifyHasError($"{ShortOption} {invalidUrl}");
```
erdembayar, line 40: Should we test if outputting expected error message something like 

`Assert.Equal("Unsupported value for --url. The value must be an absolute HTTP or HTTPS URL."?, result.Errors.First()); `

dtivel: @erdembayar, I created a [separate issue](https://github.com/dotnet/sign/issues/544) to track that feedback.  I don't think it should hold up this PR, as I believe that the change, if applied, should be applied to all tests, not just new tests.

erdembayar (approved):
LGTM

PR #541, opened 2023-01-05
CLI specification

clairernovotny:


docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+### Inputs
+
+The parameters to the signing client are as follows:
+
+`sign code AzureKeyVault [options] <files>`
+
+Options:
+
+- `-o | --output` - The output file. If omitted, overwrites input. Must be a directory if multiple files are specified.
+- `-b | --base-directory` - Base directory for files to override the working directory.
```
erdembayar, line 12: Could you be able to elaborate more what is this base directory? Is it related to `files`?

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+- `-f | --force` - Overwrites a signature if it exists.
```
erdembayar, line 13: If I understood correctly then we can sign same nupkg with multiple certificates, for example author cert, repo cert etc. Could we still do it here? 
With this option flag what exactly happens here?

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+- `-m | --max-concurrency` - Maximum concurrency (default is 4)
```
erdembayar, line 14: Does this concurrency apply to connection to Azure Vault or timestamp server?

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+- `-fl | --filelist` - Path to file containing paths of files to sign within an archive
```
erdembayar, line 15: In the past we had discussion about signing globing/multiple files same time for `nuget trusted-signers`. It's of course convenient to have one but may course [unwanted issue](https://github.com/NuGet/Home/issues/10647#issuecomment-801277353). 
After discussing with @dtivel we [decided not to implement](https://github.com/NuGet/Home/issues/10647#issuecomment-801523491) globing/multi files signing. Does same concern still apply here?

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+- `-tr | --timestamp-rfc3161` - Specifies the RFC 3161 timestamp server's URL.
+- `-d | --description` - Description
+- `-u | --descriptionUrl` - Description Url
+- `-td | --timestamp-digest` - Used with the -tr switch to request a digest algorithm used by the RFC 3161 timestamp server.
+- `-kvu | --azure-key-vault-url` - The URL to an Azure Key Vault.
+- `-kvt | --azure-key-vault-tenant-id` - The Tenant Id to authenticate to the Azure Key Vault..
+- `-kvi | --azure-key-vault-client-id` - The Client ID to authenticate to the Azure Key Vault.
+- `-kvs | --azure-key-vault-client-secret` - The Client Secret to authenticate to the Azure Key Vault.
+- `-kvc | --azure-key-vault-certificate` - The name of the certificate in Azure Key Vault.
+- `-kvm | --azure-key-vault-managed-identity` - Use a Managed Identity to access Azure Key Vault.
```
erdembayar, line 26: Are we missing `-h | --help` option here?

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+
```
erdembayar, line 27: Do we need `[-v|--verbosity <LEVEL>]` option here? Maybe customer want to see more details if there is issue with signing process.

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+`sign code AzureKeyVault [options] <files>`
+
+Options:
+
+- `-o | --output` - The output file. If omitted, overwrites input. Must be a directory if multiple files are specified.
+- `-b | --base-directory` - Base directory for files to override the working directory.
+- `-f | --force` - Overwrites a signature if it exists.
+- `-m | --max-concurrency` - Maximum concurrency (default is 4)
+- `-fl | --filelist` - Path to file containing paths of files to sign within an archive
+- `-fd | --file-digest` - The digest algorithm to hash the file with.
```
erdembayar, line 16: I expected it would be enum, right? Which algorithm does it default to?
```suggestion
- `-fd | --file-digest` <HASHALGORITHM> - The digest algorithm to hash the file with.
```

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+- `-tr | --timestamp-rfc3161` - Specifies the RFC 3161 timestamp server's URL.
+- `-d | --description` - Description
+- `-u | --descriptionUrl` - Description Url
+- `-td | --timestamp-digest` - Used with the -tr switch to request a digest algorithm used by the RFC 3161 timestamp server.
```
erdembayar, line 20: This one is also expects enum value, right? If yes what is the expected values and default value?

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+**/MyOtherLibrary.dll
+```
+
+## ClickOnce
+ClickOnce files can be signed with this tool, but it requires an extra step -- you must zip up the `publish` directory containing the `setup.exe`, `foo.application` or `foo.vsto` files along with the `Application Files` directory. The `Application Files` must only have a single subdirectory (version you want to sign). Zip these and then rename the extension to `.clickonce` before submitting to the tool. Once done, you can extract the signed files wherever you'd like for publication. If the `name` parameter is supplied, it's used in the `Mage` name to update the `Product` in the manifests. If the `descriptionUrl` parameter is supplied, it's used as the `supportUrl` in the manifests.  You should also use the `filter` parameter with the file list to sign, something like this:
+
+```
+**/ProjectAddIn1.*
+**/setup.exe
+```
```
erdembayar, line 47: Could we have sample section like [dotnet nuget sign](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-sign#examples)?

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
+# Signing CLI specficiation
+
+### Inputs
+
+The parameters to the signing client are as follows:
+
+`sign code AzureKeyVault [options] <files>`
+
```
erdembayar, line 8: Could we have [Synopsis](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-sign#synopsis) section like `dotnet nuget sign`?

erdembayar (commented):
Thank you for spec, looks good start. 
Left few comments.

docs/cli-spec.md
```diff
@@ -0,0 +1,47 @@
...
+- `-kvu | --azure-key-vault-url` - The URL to an Azure Key Vault.
+- `-kvt | --azure-key-vault-tenant-id` - The Tenant Id to authenticate to the Azure Key Vault..
+- `-kvi | --azure-key-vault-client-id` - The Client ID to authenticate to the Azure Key Vault.
+- `-kvs | --azure-key-vault-client-secret` - The Client Secret to authenticate to the Azure Key Vault.
+- `-kvc | --azure-key-vault-certificate` - The name of the certificate in Azure Key Vault.
+- `-kvm | --azure-key-vault-managed-identity` - Use a Managed Identity to access Azure Key Vault.
+
+Signing an archive type (`.zip`, `.nupkg`, `.vsix`) will open up the archive and sign any supported file types. It is strongly recommended to use the `filelist` parameter to explicitly list the files inside the archive that should be signed. Signing is recursive; it will sign contents of any detectected `Zip`, `NuPkg` or `VSIX` files inside the uploaded one. After signing contents of the archive, the archive itself is signed if supported (currently `VSIX`).
+
+### File list
```
erdembayar, line 30: This is argument, right? Could we create Argument section like https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-sign#arguments?

dtivel:
@erdembayar, thanks for the feedback.  It will not be overlooked.

I'm going to move this PR to draft state, as I think @clairernovotny mistakenly published it too early.

docs/cli-spec.md
clairernovotny: it's for any parameter that takes a relative path for files

docs/cli-spec.md
clairernovotny: This simply replaces the signature with a new one. It does not append.

docs/cli-spec.md
clairernovotny: it's in areas of the code that process multiple files at the same time -- i.e., looping over all files, reading, signing, calling api's. It's for a `Parallel.ForEach` effectively.

docs/cli-spec.md
clairernovotny: Not at this time as we must support all of the existing functionality of the legacy sign client, and this is one of the features.

docs/cli-spec.md
clairernovotny: Could be missing in the docs; I believe it's there by default with system.commandline

clairernovotny:
I'd like to get this merged asap so there's some docs on the cli usage. We can continue to update the formatting and contents of this doc after.

erdembayar:
> I'd like to get this merged asap so there's some docs on the cli usage. We can continue to update the formatting and contents of this doc after.

Could you please create tracking issue for improvement? Then I can approve it.

docs/cli-spec.md
michael-hawker: Would be good to know what the options are here for this value to provide as generally the tool is run on the CI, so getting help from the tool directly involves more steps (and it's not as discernable from the code as it's a .NET enum value)

PR #548, opened 2023-01-11
Add documentation comparing NuGet and Sign CLI

dtivel:
CC @clairernovotny

### DIFF
--- a/docs/comparisons.md
+++ b/docs/comparisons.md
@@ -33,7 +33,7 @@ macOS | ❌ | ✔️ | ❌
 Requirement | NuGet CLI | dotnet CLI | Sign CLI
 -- | -- | -- | --
 .NET Framework | ✔️ (>= 4.7.2) | ❌ | ❌
-.NET SDK | ❌ | ✔️ (>= 7) | ❌
+.NET SDK | ❌ | ✔️ (>= 5 on Windows, >= 7 on Linux, N/A on macOS) | ❌
 .NET Runtime | ❌ | ❌ | ✔️ (>= 6)
 
 ## References
--- a/docs/comparisons.md
+++ b/docs/comparisons.md
@@ -1,0 +1,41 @@
+# Signing Comparisons
+
+## NuGet
+
+The following tables summarize differences between NuGet, dotnet, and Sign CLI's. 
+
+### Features
+
+Feature | NuGet CLI | dotnet CLI | Sign CLI
+-- | -- | -- | --
+Use signing certificate from the file system | ✔️ | ✔️ | ❌
+Use signing certificate from a local store | ✔️ | ✔️ | ❌
+Use signing certificate from Azure Key Vault | ❌ | ❌ | ✔️
+Identify signing certificate by fingerprint | ✔️ | ✔️ | ❌
+Identify signing certificate by subject name | ✔️ | ✔️ | ❌
+Identify signing certificate by name (user-defined) | ❌ | ❌ | ✔️
+Can skip timestamping | ✔️ | ✔️ | ❌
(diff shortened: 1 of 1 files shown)


docs/comparisons.md
```diff
@@ -0,0 +1,41 @@
...
+Feature | NuGet CLI | dotnet CLI | Sign CLI
+-- | -- | -- | --
+Use signing certificate from the file system | ✔️ | ✔️ | ❌
+Use signing certificate from a local store | ✔️ | ✔️ | ❌
+Use signing certificate from Azure Key Vault | ❌ | ❌ | ✔️
+Identify signing certificate by fingerprint | ✔️ | ✔️ | ❌
+Identify signing certificate by subject name | ✔️ | ✔️ | ❌
+Identify signing certificate by name (user-defined) | ❌ | ❌ | ✔️
+Can skip timestamping | ✔️ | ✔️ | ❌
+Opt-in required to overwrite already signed package | ✔️ | ✔️ | ❌
```
erdembayar, line 18: According to [cli spec draft](https://github.com/dotnet/sign/pull/541/files), `dotnet sign` can overwrite already signed package if I'm reading it correctly.
image

docs/comparisons.md
```diff
@@ -0,0 +1,41 @@
...
+Windows ARM64 | ❌ | ✔️ | ❌
+Linux | ❌ | ✔️ | ❌
+macOS | ❌ | ✔️ | ❌
+
+### Requirements
+
+Requirement | NuGet CLI | dotnet CLI | Sign CLI
+-- | -- | -- | --
+.NET Framework | ✔️ (>= 4.7.2) | ❌ | ❌
+.NET SDK | ❌ | ✔️ (>= 7) | ❌
```
erdembayar, line 36: Technically `dotnet cli` can sign packages starting with `5.x.x`, right?

docs/comparisons.md
dtivel: You are correct, and that's what this doc says.  NuGet and dotnet CLI's allow you to overwrite a signature, but you have to opt in by passing `-overwrite` (NuGet CLI) or `--overwrite` (dotnet CLI).  Overwriting isn't enabled by default for NuGet and dotnet CLI's.  However, Sign CLI overwrites by default with no opt out.

docs/comparisons.md
dtivel: It's more complicated than that.  I was trying to keep it simple.  I'll update with more detail.

docs/comparisons.md
erdembayar: If `Sign CLI` overwrites by default, then do we need `-f | --force` option? Or is it for something else?

dtivel: Currently, for scoping reasons, Sign CLI is supposed to match behavior of its predecessor as much as possible.  The old SignService [overwrote signatures by default with no opt-out](https://github.com/dotnet/sign/blob/34f0acefee3673ba192ae0f72e39d029d4d2ef78/src/SignService/SigningTools/NuGetSignService.cs#L126) and [so does Sign CLI](https://github.com/dotnet/sign/blob/b85082906e9c4706efb7580009fe9ebbc320a335/src/Sign.Core/Tools/NuGetSignTool.cs#L45).

This document merely documents current behavior.

I understand and agree with the spirit of a `-overwrite` or `--force` option, but Sign CLI scenarios are more complicated than NuGet's.  With NuGet, you're overwriting a primary signature.  With Sign CLI, you're potentially overwriting different signatures of different types (e.g.:  NuGet, Authenticode, ...).  You may want to _add_ another Authenticode signature but overwrite a NuGet package signature; a simple CLI option won't cut it.

skofman1 (approved):

Issue #544, opened 2023-01-09
Test: verify error messages


dtivel:
> Should we test if it's outputting expected error message something like
>
> `Assert.Equal("Unsupported value for --url. The value must be an absolute HTTP or HTTPS URL."?, result.Errors.First()); `

_Originally posted by @erdembayar in https://github.com/dotnet/sign/pull/540#discussion_r1063917895_

PR #549, opened 2023-01-12
Test:  improve tests by checking error messages

dtivel:
Resolve https://github.com/dotnet/sign/issues/544.

This change simplifies tests a bit and checks error messages.

CC @erdembayar, @clairernovotny

### DIFF
--- a/test/Sign.Cli.Test/Options/MaxConcurrencyOptionTests.cs
+++ b/test/Sign.Cli.Test/Options/MaxConcurrencyOptionTests.cs
@@ -9,14 +9,16 @@ namespace Sign.Cli.Test
     public class MaxConcurrencyOptionTests : Int32OptionTests
     {
         public MaxConcurrencyOptionTests()
-            : base(new CodeCommand().MaxConcurrencyOption, "-m", "--max-concurrency", isRequired: false)
+            : base(new CodeCommand().MaxConcurrencyOption, "-m", "--max-concurrency")
         {
         }
 
         [Fact]
         public void Option_WhenValueFailsToParse_HasError()
         {
-            VerifyHasError("x");
+            const string value = "x";
+
+            VerifyHasErrors(value, GetUnrecognizedCommandOrArgumentMessage(value));
         }
 
         [Fact]
@@ -33,7 +35,9 @@ namespace Sign.Cli.Test
         [InlineData(-1)]
         public void Option_WhenValueIsLessThanOne_HasError(int value)
         {
-            VerifyHasError($"{LongOption} {value}");
+            VerifyHasErrors(
+                $"{LongOption} {value}",
+                GetFormattedResourceString(Resources.InvalidMaxConcurrencyValue, LongOption));
         }
     }
 }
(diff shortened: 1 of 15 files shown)


dtivel:
@erdembayar, can you take a look?

erdembayar:
> @erdembayar, can you take a look?

Sure, I'll do today.

test/Sign.Cli.Test/Options/OptionTests.cs
```diff
@@ -100,20 +111,29 @@ protected virtual void VerifyEqual(T? expectedValue, T? actualValue)
             Assert.Equal(expectedValue, actualValue);
         }
 
-        protected void VerifyHasError(string commandLine)
+        protected void VerifyHasErrors(string commandLine, params string[] expectedErrorMessages)
         {
             ParseResult result = Parse(commandLine);
+            HashSet<string> expectedMessages = new(expectedErrorMessages, StringComparer.Ordinal);
```
erdembayar, line 117: Is there any case where ordering of error message matters? If yes, then we're losing ordering here.

test/Sign.Cli.Test/Options/MaxConcurrencyOptionTests.cs
```diff
@@ -33,7 +35,9 @@ public void Option_WhenOptionIsMissing_HasDefaultValue()
         [InlineData(-1)]
         public void Option_WhenValueIsLessThanOne_HasError(int value)
```
erdembayar, line 36: Is there any limit for number of concurrent connections value? 
Anything more than 64 is not practical, right?

test/Sign.Cli.Test/Options/UriOptionTests.cs
```diff
@@ -10,15 +10,29 @@ public abstract class UriOptionTests : OptionTests<Uri?>
...
         {
         }
 
         [Fact]
         public void Option_WhenValueFailsToParse_HasError()
         {
-            VerifyHasError("3");
+            const string value = "3";
+
+            if (Option.IsRequired)
```
erdembayar, line 23: Just curious how this `IsRequired` option gets `true`, `false` values? Who is passing it?

erdembayar (approved):

test/Sign.Cli.Test/Options/OptionTests.cs
dtivel: Nope, ordering doesn't really matter.

test/Sign.Cli.Test/Options/UriOptionTests.cs
dtivel: `IsRequired` is set in the product and passed with the `Option` instance to the test.

https://github.com/dotnet/sign/blob/7b32935c42c720d96ceee799980a7f5ec2d2c296/src/Sign.Cli/CodeCommand.cs#L30-L32

test/Sign.Cli.Test/Options/MaxConcurrencyOptionTests.cs
dtivel: Where did you get 64?  An upper bound doesn't appear to be officially documented, so I wouldn't want to hard-code a limit here.

erdembayar: I just throwed some random high numbers, maybe not important.

Issue #559, opened 2023-01-14
`--verbosity` option description incorrect


dtivel:
Partial `-?` output:

```
  -v, --verbosity                                        Sets the verbosity level. Allowed values are q[uiet],
  <Critical|Debug|Error|Information|None|Trace|Warning>  m[inimal], n[ormal], d[etailed], and diag[nostic]. [default:
                                                         Warning]
```

The option is [defined as `Option<LogLevel>`](https://github.com/dotnet/sign/blob/7b32935c42c720d96ceee799980a7f5ec2d2c296/src/Sign.Cli/CodeCommand.cs#L25), where `LogLevel` is the [`Microsoft.Extensions.Logging.LogLevel` enum](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=dotnet-plat-ext-7.0).

The correct description should be:

```
  -v, --verbosity                                        Sets the verbosity level. Allowed values are None, Critical, 
  <Critical|Debug|Error|Information|None|Trace|Warning>  Error, Warning, Information, Debug, and Trace. [default: 
                                                         Warning]
```

CC @clairernovotny

PR #560, opened 2023-01-14
Update verbosity description

dtivel:
Resolve https://github.com/dotnet/sign/issues/559.

This change updates the description for `--verbosity` option to display the correct allowed values.

CC @clairernovotny

### DIFF
--- a/src/Sign.Cli/Resources.resx
+++ b/src/Sign.Cli/Resources.resx
@@ -174,8 +174,8 @@
     <comment>{Locked="RFC 3161"} is an Internet standard (https://www.rfc-editor.org/info/rfc3161) and should not be localized.</comment>
   </data>
   <data name="VerbosityOptionDescription" xml:space="preserve">
-    <value>Sets the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].</value>
-    <comment>{Locked="q[uiet]","m[inimal]","n[ormal]","d[etailed]","diag[nostic]"} are option values and should not be localized.</comment>
+    <value>Sets the verbosity level. Allowed values are None, Critical, Error, Warning, Information, Debug, and Trace.</value>
+    <comment>{Locked="None","Critical","Error","Warning","Information","Debug","Trace"} are option values and should not be localized.</comment>
   </data>
   <data name="x86NotSupported" xml:space="preserve">
     <value>Only Windows x64 is supported at this time. See https://github.com/dotnet/sign/issues/474 regarding Windows x86 support.</value>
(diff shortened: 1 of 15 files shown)


src/Sign.Cli/Resources.resx
```diff
@@ -170,7 +170,7 @@
     <value>RFC 3161 timestamp server URL.</value>
   </data>
   <data name="VerbosityOptionDescription" xml:space="preserve">
-    <value>Sets the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].</value>
+    <value>Sets the verbosity level. Allowed values are None, Critical, Error, Warning, Information, Debug, and Trace.</value>
```
erdembayar, line 173: Why are values starting with upper case? Also why not offering abbreviation? like ` -n | none`

```suggestion
<value>Sets the verbosity level. Allowed values are 'none', 'critical', 'error', 'warning', 'information', 'debug', and 'trace'.</value>
```

dtivel: I made them lower case to be consistent with other option values (like `sha256`).  I also quoted other option values elsewhere (e.g.:  `sha256`).

The original verbosity option description, which I'm updating in this PR, I stole from dotnet.exe.  However, I didn't get around to implementing it that way.  Instead, I used logging as defined in System.CommandLine.  This is why the option values changed.  It is also why abbreviations are not offered, because it was never a feature to begin with, just an incorrect description.  We don't use option value abbreviations elsewhere, and after chatting with @clairernovotny on this, it's a new feature request.

erdembayar (approved):
One none blocking comment, just curious where is the table for how different option changes logging behaviors.

src/Sign.Cli/Resources.resx
```diff
@@ -174,8 +174,8 @@
     <comment>{Locked="RFC 3161"} is an Internet standard (https://www.rfc-editor.org/info/rfc3161) and should not be localized.</comment>
   </data>
   <data name="VerbosityOptionDescription" xml:space="preserve">
-    <value>Sets the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].</value>
-    <comment>{Locked="q[uiet]","m[inimal]","n[ormal]","d[etailed]","diag[nostic]"} are option values and should not be localized.</comment>
+    <value>Sets the verbosity level. Allowed values are 'none', 'critical', 'error', 'warning', 'information', 'debug', and 'trace'.</value>
```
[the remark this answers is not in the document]
dtivel: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=dotnet-plat-ext-7.0#fields

I opened https://github.com/dotnet/sign/issues/589.

Issue #567, opened 2023-01-17
ClickOnce:  enable setting the publisher name


dtivel:
Currently, Sign CLI enables a user to set the deployment name in a ClickOnce application manifest using the `-pn` | `--publisher-name` option.  [This corresponds to the `name` attribute on the `assemblyIdentity` element.](https://learn.microsoft.com/en-us/visualstudio/deployment/assemblyidentity-element-clickonce-deployment?view=vs-2022#elements-and-attributes)  The default value is the application manifest file name.

Sign CLI does not enable a user to set the publisher name.  [This corresponds to the `publisher` attribute on the `description` element.](https://learn.microsoft.com/en-us/visualstudio/deployment/description-element-clickonce-deployment?view=vs-2022#elements-and-attributes)  [Sign CLI uses the signing certificate's CN value for the default value.](https://github.com/dotnet/sign/blob/7b32935c42c720d96ceee799980a7f5ec2d2c296/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs#L141-L145)

I propose renaming the existing `-pn` | `--publisher-name` option to `-dn` | `--deployment-name` and then introducing a new `-pn` | `--publisher-name` option to provide a value for the `publisher` attribute on the `description` element.

This should fix https://github.com/dotnet/sign/issues/412 in Sign CLI.

CC @clairernovotny

PR #568, opened 2023-01-17
Enable setting the publisher name and change old `--publisher-name` option to `--application-name`

dtivel:
Resolve https://github.com/dotnet/sign/issues/567.

This change renames the existing `-pn` | `--publisher-name` option to `-dn` | `--deployment-name`.  This option was already setting the deployment name.  Also, this change introduces a new `-pn` | `--publisher-name` option for setting publisher name. 

CC @clairernovotny

### DIFF
--- a/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
+++ b/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
@@ -64,9 +64,9 @@ namespace Sign.Core
             Logger.LogInformation(Resources.ClickOnceSignatureProviderSigning, files.Count());
 
             var args = "-a sha256RSA";
-            if (!string.IsNullOrWhiteSpace(options.DeploymentName))
+            if (!string.IsNullOrWhiteSpace(options.ApplicationName))
             {
-                args += $@" -n ""{options.DeploymentName}""";
+                args += $@" -n ""{options.ApplicationName}""";
             }
 
             Uri? timeStampUrl = options.TimestampService;
@@ -129,7 +129,7 @@ namespace Sign.Core
                             throw new Exception(message);
                         }
 
-                        string publisherParam = "";
+                        string publisherParam = string.Empty;
 
                         if (string.IsNullOrEmpty(options.PublisherName))
                         {
@@ -140,7 +140,7 @@ namespace Sign.Core
                             string publisherName = publisherIdentity!.Attribute("name")!.Value;
                             Dictionary<string, List<string>> dict = DistinguishedNameParser.Parse(publisherName);
  
-                            if (dict.TryGetValue("CN", out var cns))
+                            if (dict.TryGetValue("CN", out List<string>? cns))
                             {
                                 // get the CN. it may be quoted
                                 publisherParam = $@"-pub ""{string.Join("+", cns.Select(s => s.Replace("\"", "")))}"" ";
--- a/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
+++ b/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
@@ -64,9 +64,9 @@ namespace Sign.Core
             Logger.LogInformation(Resources.ClickOnceSignatureProviderSigning, files.Count());
 
             var args = "-a sha256RSA";
-            if (!string.IsNullOrWhiteSpace(options.PublisherName))
+            if (!string.IsNullOrWhiteSpace(options.DeploymentName))
             {
-                args += $@" -n ""{options.PublisherName}""";
+                args += $@" -n ""{options.DeploymentName}""";
             }
 
             Uri? timeStampUrl = options.TimestampService;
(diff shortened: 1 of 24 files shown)


src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
```diff
@@ -129,19 +129,26 @@ public async Task SignAsync(IEnumerable<FileInfo> files, SignOptions options)
                             throw new Exception(message);
                         }
 
-                        // Read the publisher name from the manifest for use below
-                        XDocument manifestDoc = XDocument.Load(manifestFile.FullName);
-                        XNamespace ns = manifestDoc.Root!.GetDefaultNamespace();
-                        XElement? publisherIdentity = manifestDoc.Root.Element(ns + "publisherIdentity");
-                        string publisherName = publisherIdentity!.Attribute("name")!.Value;
+                        string publisherParam = "";
```
erdembayar, line 132:
```suggestion
                        string publisherParam = string.Empty;
```

src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
```diff
@@ -129,19 +129,26 @@ public async Task SignAsync(IEnumerable<FileInfo> files, SignOptions options)
...
-
-                        Dictionary<string, List<string>> dict = DistinguishedNameParser.Parse(publisherName);
-                        if (dict.TryGetValue("CN", out var cns))
+                        if (string.IsNullOrEmpty(options.PublisherName))
+                        {
+                            // Read the publisher name from the manifest for use below
+                            XDocument manifestDoc = XDocument.Load(manifestFile.FullName);
+                            XNamespace ns = manifestDoc.Root!.GetDefaultNamespace();
+                            XElement? publisherIdentity = manifestDoc.Root.Element(ns + "publisherIdentity");
+                            string publisherName = publisherIdentity!.Attribute("name")!.Value;
```
erdembayar, line 140: Should we replace `"name"` string with `const` on top?

src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
```diff
@@ -129,19 +129,26 @@ public async Task SignAsync(IEnumerable<FileInfo> files, SignOptions options)
...
+                            Dictionary<string, List<string>> dict = DistinguishedNameParser.Parse(publisherName);
+ 
+                            if (dict.TryGetValue("CN", out var cns))
```
erdembayar, line 143: Same here for `"CN"`, probably better to create const for it.

src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
```diff
@@ -129,19 +129,26 @@ public async Task SignAsync(IEnumerable<FileInfo> files, SignOptions options)
...
+                            {
+                                // get the CN. it may be quoted
+                                publisherParam = $@"-pub ""{string.Join("+", cns.Select(s => s.Replace("\"", "")))}"" ";
```
erdembayar, line 146: Do we have test for this code path?

dtivel:
One consideration is changing `-dn` | `--deployment-name` to `-an` | `--application-name`.

[This](https://learn.microsoft.com/en-us/visualstudio/deployment/assemblyidentity-element-clickonce-deployment?view=vs-2022#elements-and-attributes) says the `assemblyIdentity.name` value "[i]dentifies the human-readable name of the deployment for informational purposes."

mage.exe CLI's help says the value of the corresponding option (`-n` | `-Name`) "[s]pecifies the name of the application whose manifest is being generated or updated."

[mage.exe documentation](https://learn.microsoft.com/en-us/dotnet/framework/tools/mage-exe-manifest-generation-and-editing-tool#new-and-update-command-options) says the `-n` | `-Name` option "is used to identify the application."

erdembayar (commented):
Few comments.

src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
dtivel: A good rule of thumb for this is if the string literal is only used in one place, it's optional.  I didn't use a constant here because `"name"` is only used once, and introducing a constant is of debatable value.  However, as soon as a string literal is used in multiple places, a constant should be used.

src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
dtivel: See my previous comment.

src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
dtivel: Yep, right [here](https://github.com/dotnet/sign/pull/568/files#diff-8c821ccfd0139e3ab038bc0f20d592d1899314a0b7be71caf9e63860fa1f94f8R199).

erdembayar:
@dtivel 
Could you please update PR description? I'm confused what is the exact change here?
According to desc we're renaming `-pn | --publisher-name` option to `-dn | --deployment-name`, yet we're re-introducing same `-pn | --publisher-name` option we just renamed, bit confusing to digest. For me PR looks just adding `-an | --application-name` option.

erdembayar (approved):
LGTM. Probably need to update the PR description, current state of PR bit different from PR desc.

dtivel:
Thanks for the review.

Think of this PR as repurposing the existing `-pn` | `--publisher-name` option to set publisher name --- since the previous implementation didn't really do that --- and introducing a new option to set application name.

Issue #547, opened 2023-01-11
Cannot overwrite signature on already signed NuGet package


dtivel:

clairernovotny:
Isn't there a --force parameter? Does that work or is it broken?

dtivel:
Yes, [the draft CLI spec proposes a `--force` parameter](https://github.com/dotnet/sign/blob/dtivel/cli-spec/docs/cli-spec.md); however:

* The option isn't currently implemented.
* The preview Sign CLI is supposed to match behavior of its predecessor. The old SignService [overwrote signatures by default with no opt-out](https://github.com/dotnet/sign/blob/34f0acefee3673ba192ae0f72e39d029d4d2ef78/src/SignService/SigningTools/NuGetSignService.cs#L126) and [so does Sign CLI](https://github.com/dotnet/sign/blob/b85082906e9c4706efb7580009fe9ebbc320a335/src/Sign.Core/Tools/NuGetSignTool.cs#L45).
* The default, which is to overwrite, is broken.

PR #569, opened 2023-01-18
Enable signing already signed NuGet package

dtivel:
Resolve https://github.com/dotnet/sign/issues/547.

This change enables signing an already signed NuGet package.

CC @clairernovotny

### DIFF
--- a/test/Sign.Core.Test/TestInfrastructure/DirectoryServiceStub.cs
+++ b/test/Sign.Core.Test/TestInfrastructure/DirectoryServiceStub.cs
@@ -1,0 +1,47 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+namespace Sign.Core.Test
+{
+    internal sealed class DirectoryServiceStub : IDirectoryService
+    {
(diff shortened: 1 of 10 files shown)


test/Sign.Core.Test/TestInfrastructure/DirectoryServiceStub.cs
```diff
@@ -0,0 +1,47 @@
...
+        internal DirectoryServiceStub()
+        {
+            Directories = _directories = new List<DirectoryInfo>();
+        }
+
+        public DirectoryInfo CreateTemporaryDirectory()
+        {
+            DirectoryInfo directory = new(Path.Combine(Path.GetTempPath(), Path.GetRandomFileName()));
+
+            directory.Create();
```
erdembayar, line 22: Is there any slim chance of directory with same name already exist by random chance?
Should we have retried logic just in case?

dtivel: There is a chance, but it seems extraordinarily unlikely.  Although adding a retry would make it even less unlikely, just how unlikely is acceptable?  The amount of time to write, test, review, etc. this code doesn't seem worth it given the already low chances of this being a problem.

erdembayar (approved):

PR #572, opened 2023-01-19
Add sample pipelines and getting started

clairernovotny:


dtivel (approved):

skofman1:
Copying the comment from chat:
For the readme, the question is "who is the audience?". Is this someone who is code signing for the first time, signing for the first time using KV, or an experienced user who is migrating from the "Sign service" to "Sign CLI". For "first timers", I would start with "self-signed certificate" vs. "EV certificate" (and explain what to choose for different scenarios). Also, in the "Azure Setup" section you could link to instructions on how to create a KV, how to upload a cert / create a self-signed cert in KV) and mention that it must be Premium SKU for EV certs (because this might be a gotcha for folks).

clairernovotny:
Right now it's for existing users. I'd like to avoid getting into the details of code signing certificate types in this document.

skofman1:
Makes sense. Can we track this in an issue for Public preview?

PR #598, opened 2023-02-27
Make ACS the default timestamping service

dtivel:
Resolve https://github.com/dotnet/sign/issues/597.

This change makes Azure Code Signing timestamping service the default.  The default can be overridden with the `-t` | `--timestamp-url` option.

Note that changing `IsRequired` from `true` to `false` means that a user isn't required to pass the option.  The new default value guarantees that there will be a value, so timestamping is still effectively required.

CC @clairernovotny

### DIFF
--- a/src/Sign.Cli/CodeCommand.cs
+++ b/src/Sign.Cli/CodeCommand.cs
@@ -30,11 +30,11 @@ namespace Sign.Cli
         {
             DescriptionOption.IsRequired = true;
             DescriptionUrlOption.IsRequired = true;
-            TimestampUrlOption.IsRequired = true;
 
             MaxConcurrencyOption.SetDefaultValue(4);
             FileDigestOption.SetDefaultValue(HashAlgorithmName.SHA256);
             TimestampDigestOption.SetDefaultValue(HashAlgorithmName.SHA256);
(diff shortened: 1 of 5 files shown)


src/Sign.Cli/CodeCommand.cs
```diff
@@ -30,11 +30,11 @@ internal CodeCommand()
         {
             DescriptionOption.IsRequired = true;
             DescriptionUrlOption.IsRequired = true;
-            TimestampUrlOption.IsRequired = true;
 
             MaxConcurrencyOption.SetDefaultValue(4);
             FileDigestOption.SetDefaultValue(HashAlgorithmName.SHA256);
             TimestampDigestOption.SetDefaultValue(HashAlgorithmName.SHA256);
+            TimestampUrlOption.SetDefaultValue(new Uri("http://timestamp.acs.microsoft.com"));
```
erdembayar, line 37: Is this expected to be stable service after many years?  If I understood correctly actual certificates usually expire after few years, then this one is kick in,

dtivel: It is expected to be stable.  The URL is separate from which signing certificate(s) are used.  The signing certificates rotate over time, while the URL should be stable.

erdembayar (approved):

Issue #606, opened 2023-03-02
Loc comments incorrect/lacking


dtivel:
Some resource strings cannot be localized because the comments are incorrect and/or lacking.  Each resource string with a replacement token should have good functional and contextual commenting.

PR #607, opened 2023-03-02
Update resource strings and comments

dtivel:
Resolve https://github.com/dotnet/sign/issues/606.

This change updates resource string comments (and a couple resource strings) to have correct functional and contextual commenting.  This is necessary for localization.

CC @clairernovotny

### DIFF
--- a/src/Sign.Core/xlf/Resources.fr.xlf
+++ b/src/Sign.Core/xlf/Resources.fr.xlf
@@ -4,8 +4,8 @@
     <body>
       <trans-unit id="AzureSignToolSignatureProviderSigning">
         <source>Signing SignTool job with {count} files.</source>
-        <target state="translated">Signature du travail SignTool avec {count} fichiers.</target>
-        <note>{count} is the number of files to be signed.  The literal text inside the brackets should not be localized.</note>
+        <target state="needs-review-translation">Signature du travail SignTool avec {count} fichiers.</target>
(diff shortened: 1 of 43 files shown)


src/Sign.Core/xlf/Resources.fr.xlf
```diff
@@ -159,8 +159,8 @@
       </trans-unit>
       <trans-unit id="VsixSignatureProviderSigning">
         <source>Signing OpenVsixSignTool job with {count} files.</source>
-        <target state="translated">Signature du travail OpenVsixSignTool avec {count} fichiers.</target>
-        <note>{count} is the number of files to be signed.  The literal text inside the brackets should not be localized.</note>
+        <target state="needs-review-translation">Signature du travail OpenVsixSignTool avec {count} fichiers.</target>
+        <note>{Placeholder="{count}"} is the number of files to be signed.</note>
```
erdembayar, line 163: Should we keep the `The literal text inside the brackets should not be localized.` part here?

dtivel: Nope, it's not necessary.

dtivel: (`{Placeholder="..."}` already tells the loc engine that the text inside the brackets should not be localized.)

erdembayar (approved):

PR #617, opened 2023-03-31
Add more tests

dtivel:
Resolve https://github.com/dotnet/sign/issues/616.

This change adds:

* globbing tests (for both the file argument and in file list option)
* signing tests for:
   * portable executable file (.dll)
   * PowerShell script (.ps1)
   * MSIX bundle (.msixbundle)
   * MSIX (.msix)
   * VSIX (.vsix)

The signing tests enable local signing without Azure Key Vault by substituting the default `IKeyVaultService` implementation with a test stub.  Verification checks signature integrity and signing certificates but not trust., because the local test certificate is not trusted.

To enable these tests, I needed to introduce `IServiceProviderFactory` and a test timestamping service.

CC @clairernovotny


erdembayar (approved):

dtivel:
@Thanks, @erdembayar

Issue #618, opened 2023-03-31
Set default input/output encoding to UTF-8


dtivel:
Currently, console input/output encoding is whatever the console's default is.  If the default encoding does not support Unicode and if data displayed contains Unicode characters, then characters will be garbled.

By temporarily setting the input/output encoding to UTF-8, output will not be garbled.

CC @clairernovotny

PR #619, opened 2023-03-31
Set default input/output encoding to UTF-8

dtivel:
Resolve https://github.com/dotnet/sign/issues/618.

Temporarily set the default input/output encoding to UTF-8 to ensure Unicode characters aren't garbled.

CC @clairernovotny


src/Sign.Cli/TemporaryConsoleEncoding.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        private readonly Encoding _defaultInputEncoding;
+        private readonly Encoding _defaultOutputEncoding;
+
+        internal TemporaryConsoleEncoding()
+        {
+            _defaultInputEncoding = Console.InputEncoding;
+            _defaultOutputEncoding = Console.OutputEncoding;
+
+            Console.InputEncoding = Encoding.UTF8;
```
erdembayar, line 19: Should we have escape hatch something like env var which default to previous experience just in case?
Encoding thing could be tricky.

erdembayar (approved):

Issue #630, opened 2023-05-12
Sign CLI doesn't support ClickOnce RFC 3161 timestamping


dtivel:
This project uses a copy of mansign2.cs from .NET Reference Source.  However, the latest version is in the dotnet/msbuild repository.

Crucially, the .NET Reference Source version performs only [legacy Authenticode timestamping](
https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/inc/mansign2.cs#L1506), while the dotnet/msbuild version first attempts [RFC 3161 timestamping](https://github.com/dotnet/msbuild/blob/9deb5b58fb2b02990cba4b7add8c8dcbb9f658ed/src/Tasks/ManifestUtil/mansign2.cs#L886) before [falling back to legacy Authenticode timestamping](https://github.com/dotnet/msbuild/blob/9deb5b58fb2b02990cba4b7add8c8dcbb9f658ed/src/Tasks/ManifestUtil/mansign2.cs#L910).

[This change](https://github.com/dotnet/msbuild/commit/0551aa0c27b7d1b371ada34c65f6fdb56ba2a9b8) never made it back into .NET Reference Source.

CC @clairernovotny

PR #634, opened 2023-05-22
ClickOnce:  enable RFC 3161 timestamping

dtivel:
Resolve https://github.com/dotnet/sign/issues/630.

This change updates Sign CLI's copy of Mage-related files to the latest version in dotnet/msbuild.  This brings in RFC 3161 timestamping support.

CC @clairernovotny


src/Sign.Core/GlobalSuppressions.cs
```diff
@@ -0,0 +1,8 @@
+﻿#pragma warning disable IDE0073 // The file header does not match the required text
+// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+using System.Diagnostics.CodeAnalysis;
+[assembly: SuppressMessage("Interoperability", "CA1416:Validate platform compatibility", Justification = "This program onl supports Windows at this time.", Scope = "member", Target = "~M:Sign.Core.ManifestSigner.Sign(System.IO.FileInfo,System.Security.Cryptography.X509Certificates.X509Certificate2,System.Security.Cryptography.RSA,Sign.Core.SignOptions)")]
```
joelverhagen, line 8:
```suggestion
[assembly: SuppressMessage("Interoperability", "CA1416:Validate platform compatibility", Justification = "This program only supports Windows at this time.", Scope = "member", Target = "~M:Sign.Core.ManifestSigner.Sign(System.IO.FileInfo,System.Security.Cryptography.X509Certificates.X509Certificate2,System.Security.Cryptography.RSA,Sign.Core.SignOptions)")]
```

test/Sign.Core.Test/TestInfrastructure/TemporaryEnvironmentPath.cs
```diff
@@ -0,0 +1,30 @@
...
+            string newPaths = string.Join(Path.PathSeparator.ToString(), paths, path);
+
+            Environment.SetEnvironmentVariable(Name, newPaths);
+        }
+
+        public void Dispose()
+        {
+            Environment.SetEnvironmentVariable(Name, _originalEnvironmentPath);
+
+            GC.SuppressFinalize(this);
```
joelverhagen, line 27: I've always acted under the guidance from this old StackOverflow answer -- that `GC.SuppressFinalize` should only be used if there's a finalizer: https://stackoverflow.com/a/151244. Could you help me understand the need for this line?

I understand that you want to leave the process env in it's original state, but it seems like it will only ever be best effort (caller invokes `Dispose` as expected) or unnecessary (process terminates without `Dispose` being called in which case calling `SetEnvironmentVariable` isn't needed anyway since the default overload you're calling uses `EnvironmentVariableTarget.Process` and the process's copy of env will be discarded anyway).

joelverhagen (approved):

test/Sign.Core.Test/Native/SignedCmiManifest2Tests.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+
+                byte[] bytes = GetTimestampBytes(manifest);
+
+                Assert.True(
+                    Rfc3161TimestampToken.TryDecode(
+                        bytes,
+                        out Rfc3161TimestampToken? timestampToken,
+                        out int bytesConsumed));
+                Assert.True(timestampToken.TokenInfo.HashAlgorithmId.IsEqualTo(Oids.Sha256));
```
agr, line 62: `Assert.NotNull(timestampToken)` above this line?

agr, line 62: Does `Assert.Equal` not work for `Oid`?

test/Sign.Core.Test/TestInfrastructure/TemporaryEnvironmentPath.cs
```diff
@@ -0,0 +1,30 @@
...
+        private const string Name = "PATH";
+
+        private readonly string? _originalEnvironmentPath;
+
+        internal TemporaryEnvironmentPath(string path)
+        {
+            _originalEnvironmentPath = Environment.GetEnvironmentVariable(Name);
+
+            string paths = _originalEnvironmentPath ?? string.Empty;
+            string newPaths = string.Join(Path.PathSeparator.ToString(), paths, path);
```
agr, line 18: Don't you want `path` prepended to existing one? So whatever you add is found before what already exists in `PATH` if conflict exists?

agr, line 18: Also, `string.Join` has overloads that accept `char` as its first argument, no need to call `.ToString()` here.

test/Sign.Core.Test/TestInfrastructure/TemporaryEnvironmentPath.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+namespace Sign.Core.Test
+{
+    internal sealed class TemporaryEnvironmentPath : IDisposable
```
agr, line 7: `TemporaryEnvironmentPathOverride`? I think that would better convey what the class does.

test/Sign.Core.Test/TestInfrastructure/TemporaryEnvironmentPath.cs
dtivel: Because the class is `sealed`, it's not necessary.  It's also no harm leaving it in; however, I'll remove it.

src/Sign.Core/Native/mansign2.cs
```diff
@@ -1,80 +1,248 @@
...
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE 
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER 
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, 
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE 
-// SOFTWARE.
-//
-
-// From https://github.com/Microsoft/referencesource/blob/7de0d30c7c5ef56ab60fee41fcdb50005d24979a/inc/mansign2.cs
```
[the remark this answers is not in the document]
dtivel: It's in the commit description:  https://github.com/dotnet/sign/pull/634/commits/0f364fa0799776803db45d16a1ff7a447311609a

test/Sign.Core.Test/Native/SignedCmiManifest2Tests.cs
dtivel: Not necessary.  `Rfc3161TimestampToken.TryDecode(...)` has `[NotNullWhen(true)]` on the `Rfc3161TimestampToken` parameter, so it should never be `null` by line 62, and I'm not trying to test .NET itself.

test/Sign.Core.Test/Native/SignedCmiManifest2Tests.cs
```diff
@@ -0,0 +1,100 @@
...
+                RSASignaturePadding.Pkcs1);
+            DateTimeOffset now = DateTimeOffset.Now;
+
+            return request.CreateSelfSigned(now.AddMinutes(-5), now.AddMinutes(5));
+        }
+
+        private static TemporaryEnvironmentPath CreateTemporaryEnvironmentPath()
+        {
+            string windir = Environment.GetEnvironmentVariable("windir")!;
+            string netfxDir = $@"{windir}\Microsoft.NET\Framework64\v4.0.30319";
```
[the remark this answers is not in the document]
dtivel: Nope.  Windows only.

test/Sign.Core.Test/TestInfrastructure/TemporaryEnvironmentPath.cs
dtivel: Great point.  The product itself has the same behavior.  I'll keep that behavior for now and open an issue to think about this more.

test/Sign.Core.Test/Native/SignedCmiManifest2Tests.cs
dtivel: It does not.

Issue #639, opened 2023-07-26
Enable pluggable signature providers


dtivel:
Currently, Sign CLI only supports signing with Azure Key Vault.  Sign CLI should support pluggable signature providers to enable other private key storage options.

CC @clairernovotny

mihai-tracer:
Any new updates on this?

dlemstra:
I think we can close this issue because we made changes to support this when support for Trusted Signing was added.

elachlan:
Is there any progress on this? I'd like to use it with AWS HSM.

PR #640, opened 2023-07-26
Design proposal for signature provider plugins

dtivel:
Progress on https://github.com/dotnet/sign/issues/639.

CC @clairernovotny

### DIFF
--- a/docs/design/signature-provider-plugins.md
+++ b/docs/design/signature-provider-plugins.md
@@ -1,0 +1,444 @@
+# Sign CLI Signature Provider Plugins
+
+**Owners** [Damon Tivel](https://github.com/dtivel) | [Claire Novotny](https://github.com/clairernovotny)
+
+Recent CAB Forum updates to baseline requirements<sup>[1](#r1)</sup> strengthened storage requirements for private keys of publicly trusted code signing certificates.  While older, less secure storage options (e.g.:  [PKCS #12 & PFX](https://en.wikipedia.org/wiki/PKCS_12)) became obsolete, more secure options (e.g.:  [HSM](https://en.wikipedia.org/wiki/Hardware_security_module)) became standard.
+
+As of writing this, Sign CLI only supports digest signing using Azure Key Vault.  To support users with private keys stored elsewhere (e.g.:  a different cloud provider, a signing service, or Windows' certificate store), Sign CLI needs a plugin model for signature providers.  Sign CLI users should be able to install a plugin that provides signing capabilities for their private key.
+
+To be clear, there is nothing in this proposed plugin model that would preclude a plugin author from creating a plugin that enables signing using a PFX file, and such a plugin would be welcome to a subset of users.  However, given the relative lack of support in existing signing tools for more secure private key storage options, the primary driver for this proposal is enabling support for more secure storage options.
+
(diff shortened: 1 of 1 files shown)


docs/design/signature-provider-plugins.md
```diff
@@ -0,0 +1,444 @@
...
+`isRequired` | no | `true`/`false` | `false` | `true`, `false`
+
+Example:
+
+```JSON
+{
+  "name": "azure-key-vault",
+  "description": "Use Azure Key Vault.",
+  "entryPoint": {
+    "filePath": "lib/net6.0/Microsoft.Azure.KeyVault.Sign.dll",
```
agr, line 295: Would there be a reason for the "plugin package" to target multiple frameworks? If yes, how would that `plugin.json` would look like?

docs/design/signature-provider-plugins.md
```diff
@@ -0,0 +1,444 @@
...
+    * `dataType`: The option's data type.  Lets Sign CLI know how to parse user input.
+    * `defaultValue`: The option's default value.  Only used if the effective value for `isRequired` is `false`; otherwise, this is ignored.
+    * `isRequired`: Whether the option is required or not.
+
+Parameter | Required | JSON Type | Default | Possible Values
+-- | -- | -- | -- | --
+`name` | yes | string | N/A | N/A
+`description` | yes | string | N/A | N/A
+`aliases` | yes | array of strings | N/A | N/A
+`dataType` | no | string | `Text` | `Text`, `Boolean`, `Uri`
```
agr, line 284: I imagine, plugin would have to validate parameters anyway, so why not just read and pass them to a plugin as strings, then let it sort out the conversion/interpretation?

docs/design/signature-provider-plugins.md
clairernovotny: Not sure about multiple frameworks as the tool is >.NET 6 only. I believe the AssemblyLoadContext will automatically pickup runtimes if they're in subfolders to handle native dependencies. @dtivel?

docs/design/signature-provider-plugins.md
```diff
@@ -0,0 +1,444 @@
...
+       <ItemGroup>
+           <BuiltProjectOutputGroupOutput
+               Include="$(ProjectDepsFilePath)"
+               TargetPath="$(ProjectDepsFileName)"
+               FinalOutputPath="$(ProjectDepsFilePath)" />
+       </ItemGroup>
+   </Target>
+   ```
+
+#### <a name="plugin-json-file"></a>The `plugin.json` file
```
skofman1, line 257: I wonder if the design can be simplified by moving the plugin.json content elsewhere. For example: name can be inferred from package name or from class name, or it can be part of the ISignatureProviderPlugin. Path can be inferred from looking at all the classes that implement ISignatureProviderPlugin and auto-adding them. 
The reason I'm suggesting this, is because editing an additional file can be error prone for developers and increase plugin development time.

docs/design/signature-provider-plugins.md
agr: There is a direct file link in the configuration, there doesn't seem to be any room for automatic choices.

docs/design/signature-provider-plugins.md
dtivel: Thoughts, @clairernovotny?

clairernovotny: I don't have a strong opinion on this other than having common parameters validated in the base could provide a more consistent experience.

dtivel:
Closing this in favor of https://github.com/dotnet/designs/pull/304.

CC @clairernovotny

PR #656, opened 2024-02-13
Rewrite ClickOnce signing to sign files directly instead of via the .clickonce zip mechanism

### LINKED ISSUE #643: Individual signing of clickonce files
We would like the ability to sign individual clickonce-related files separately. E.g. explicitly signing the binaries, the .manifest and then the .application files without going via the .clickonce zip package format that has been created for this tool.

The build process for our applications of course does not produce such a file automatically, nor is it usable for deployment. That means that we have to:

- create the zip explicitly in our build pipeline
- pass it to `dotnet sign`, which will then unzip it again into a temporary location, sign the relevant files, and re-zip it
- then we have to unzip it again in order to actually deploy it anywhere

This is fine for small applications, but for larger clickonce packages it's quite slow and doesn't seem worth it at all.

Therefore I'm proposing to add a new type of SignatureProvider class that natively handles .deploy, .manifest and .application files and signs them in the proper order without needing to bother with a zip file.

Would a PR implementing such a feature be accepted?

jackmtpt:
This is my first-draft implementation of signing for clickonce files that doesn't require you to zip up everything into the .clickonce file. Instead, you can just point the tool at your .application file (or .vsto) and it will figure out which other files need to be signed and sign them all for you. The actual signing still happens in a temporary path (as it does for other file types) and then the results are copied back in-place to the original location.

I wrote this a couple of months ago but finally got approval from my company to contribute it back to the project. I've rebased it onto the latest `main` this morning.

We're not using this in production yet; we'd prefer to wait until this work has been accepted into the upstream repository before we build on it internally. I'm happy to work with the maintainers and change the code/design around as-needed to get this merged.

Resolves #643 and #470.

### DIFF
--- a/src/Sign.Cli/Resources.resx
+++ b/src/Sign.Cli/Resources.resx
@@ -137,7 +137,7 @@
     <comment>{Locked="sha256", "sha384", "sha512"} are cryptographic hash algorithm names and should not be localized.</comment>
   </data>
   <data name="FileListOptionDescription" xml:space="preserve">
-    <value>Path to file containing paths of files to sign within an archive.</value>
+    <value>Path to file containing paths of files to sign within an archive or ClickOnce package.</value>
   </data>
   <data name="InvalidBaseDirectoryValue" xml:space="preserve">
     <value>Invalid value for {0}. The value must be a fully rooted directory path.</value>
--- a/src/Sign.Cli/Resources.resx
+++ b/src/Sign.Cli/Resources.resx
@@ -137,7 +137,7 @@
     <comment>{Locked="sha256", "sha384", "sha512"} are cryptographic hash algorithm names and should not be localized.</comment>
   </data>
   <data name="FileListOptionDescription" xml:space="preserve">
-    <value>Path to file containing paths of files to sign within an archive or ClickOnce package.</value>
+    <value>Path to file containing paths of files to sign or to exclude from signing.</value>
   </data>
   <data name="InvalidBaseDirectoryValue" xml:space="preserve">
     <value>Invalid value for {0}. The value must be a fully rooted directory path.</value>
--- a/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
+++ b/src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
@@ -7,6 +7,7 @@ using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 using System.Xml.Linq;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.FileSystemGlobbing.Abstractions;
 using Microsoft.Extensions.Logging;
 
 namespace Sign.Core
@@ -21,6 +22,7 @@ namespace Sign.Core
         private readonly IMageCli _mageCli;
         private readonly IManifestSigner _manifestSigner;
         private readonly ParallelOptions _parallelOptions = new() { MaxDegreeOfParallelism = 4 };
+        private readonly IFileMatcher _fileMatcher;
 
         // Dependency injection requires a public constructor.
         public ClickOnceSignatureProvider(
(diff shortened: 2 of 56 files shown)


jackmtpt:
@dotnet-policy-service agree company="Trayport"

clairernovotny (changes requested):
Thank you for this, it looks great!

It looks like support for `.clickonce` extensions was removed? I think that's ok especially as we're still in pre-release, but I think we should add an explicit error message that directs folks to the right approach. 

Can you please add a check with an error message in the application if a user tries to pass in a `.clickonce` file and also update the docs here https://github.com/dotnet/sign#clickonce? The error message can point to the docs.

jackmtpt:
Yeah I figured it was fine to just remove it all since as you say, it's still pre-release and it simplifies things considerably. I'll sort out the docs + error message now.

src/Sign.Core/SignatureProviders/ISignatureProvider.cs
```diff
@@ -8,5 +8,10 @@ internal interface ISignatureProvider
     {
         bool CanSign(FileInfo file);
         Task SignAsync(IEnumerable<FileInfo> files, SignOptions options);
+        // Some signature mechanisms (e.g. ClickOnce) require extra files alongside the main file to be signed.
+        // We can't rely on the user specifying everything (and even if we did, we sign all inputs in parallel
+        // so we'd have to add extra synchronisation) so this method instructs an implementation to grab all
+        // dependencies of a file and copy them to the specified directory.
+        void CopySigningDependencies(FileInfo file, DirectoryInfo destination, SignOptions options) { }
```
clairernovotny, line 15: Are there tests that cover usage of this to ensure it's called where it's needed? Also tests for any implementation that ensure the right files are detected and copied to the expected destination?

jackmtpt: I thought that it was covered by the existing clickonce tests but you're right, it's not since the test just constructs an already-valid directory structure. I'll add some.

jackmtpt: Pushed a commit with a test for the clickonce-specific behaviour. None of the other signature providers need it (aside from the aggregate that's just passing the call down). The `Signer` class is the only place it's called from and all the signing calls end up there eventually so that should be sufficient I think.

src/Sign.Core/SignatureProviders/ClickOnceSignatureProvider.cs
```diff
@@ -78,116 +82,114 @@ public async Task SignAsync(IEnumerable<FileInfo> files, SignOptions options)
             using (X509Certificate2 certificate = await _certificateProvider.GetCertificateAsync())
             using (RSA rsaPrivateKey = await _signatureAlgorithmProvider.GetRsaAsync())
             {
-                // This outer loop is for a .clickonce file
+                // This outer loop is for a single deployment manifest (.application/vsto).
```
dtivel, line 85: Remove `single`.  Although the caller passes one now, a caller could pass several in the future, and this code handles both cases.

src/Sign.Cli/Resources.resx
```diff
@@ -137,7 +137,7 @@
     <comment>{Locked="sha256", "sha384", "sha512"} are cryptographic hash algorithm names and should not be localized.</comment>
   </data>
   <data name="FileListOptionDescription" xml:space="preserve">
-    <value>Path to file containing paths of files to sign within an archive.</value>
+    <value>Path to file containing paths of files to sign within an archive or ClickOnce package.</value>
```
dtivel, line 140: I feel like we should be more generic, rather than be more specific.  It could also contain paths to DLL's, EXE's, NuGet packages, VSIX's, etc.

dtivel (changes requested):
@jackmtpt, first, big thanks for the contribution.  It's appreciated.

Second, I reviewed your PR, and it looks good.  Thank you for adding tests.

I had some minor feedback, which I offered up as a [commit](https://github.com/dtivel/sign/commit/7fc4eecb1797a005adb3d3cd50916880ce556cda) which you can simply incorporate.

Please rebase on latest main and ping me when ready for re-review.  Thanks!

jackmtpt:
> @jackmtpt, first, big thanks for the contribution. It's appreciated.
> 
> Second, I reviewed your PR, and it looks good. Thank you for adding tests.
> 
> I had some minor feedback, which I offered up as a [commit](https://github.com/dtivel/sign/commit/7fc4eecb1797a005adb3d3cd50916880ce556cda) which you can simply incorporate.
> 
> Please rebase on latest main and ping me when ready for re-review. Thanks!

Cheers - your changes look good and I've applied your commit. Rebased to main.

clairernovotny (approved):
LGTM. @dtivel good to go?

PR #660, opened 2024-03-01
Fix for "treat output as directory" for multiple input files

spilkercompra:
These comments in Signer.cs should be true:

// Special case if there's only one input file and the output has a value, treat it as a file
   [...]
   // See if it has a file extension and if not, treat as a directory and use the input file name
[...]
// if the output is specified, treat it as a directory, if not, overwrite the current file

### DIFF
--- a/src/Sign.Core/Signer.cs
+++ b/src/Sign.Core/Signer.cs
@@ -104,7 +104,7 @@ namespace Sign.Core
                     else
                     {
                         // if the output is specified, treat it as a directory, if not, overwrite the current file
-                        if (!string.IsNullOrWhiteSpace(outputFile))
+                        if (string.IsNullOrWhiteSpace(outputFile))
                         {
                             output = new FileInfo(input.FullName);
                         }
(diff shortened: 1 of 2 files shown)


clairernovotny (approved):

clairernovotny:
LGTM -- @dtivel?

spilkercompra:
@microsoft-github-policy-service agree

spilkercompra:
@dotnet-policy-service agree

src/Sign.Core/Signer.cs
```diff
@@ -104,7 +104,7 @@ public Signer(IServiceProvider serviceProvider, ILogger<ISigner> logger)
                     else
                     {
                         // if the output is specified, treat it as a directory, if not, overwrite the current file
-                        if (!string.IsNullOrWhiteSpace(outputFile))
+                        if (string.IsNullOrWhiteSpace(outputFile))
```
dtivel, line 107: During the conversion to a standalone client, it looks the semantics of the [original expression](https://github.com/dotnet/sign/blob/34f0acefee3673ba192ae0f72e39d029d4d2ef78/src/SignClient/SignCommand.cs#L287) weren't preserved.  Good catch.

dtivel (approved):
LGTM.  Thanks for your contribution and thanks for including tests, @spilkercompra.

Issue #667, opened 2024-03-19
Move to .NET 8.0


dtivel:
[.NET 6 runtime reaches end of support in November 2024.](https://dotnet.microsoft.com/platform/support/policy/dotnet-core)

PR #668, opened 2024-03-19
Move to .NET 8

dtivel:
This PR moves from .NET 6 to .NET 8 and updates dependencies from 7.0.x to latest release 8.0.x.

Resolve https://github.com/dotnet/sign/issues/667.


.vsts-ci.yml
```diff
@@ -104,7 +104,7 @@ stages:
               InputType: Basic
               Function: analyze
               TargetPattern: binskimPattern
-              AnalyzeTargetBinskim: $(Build.SourcesDirectory)\artifacts\bin\Sign.Cli\$(_BuildConfig)\net6.0\win10-x64\publish\*.dll
+              AnalyzeTargetBinskim: $(Build.SourcesDirectory)\artifacts\bin\Sign.Cli\$(_BuildConfig)\net8.0\win-x64\publish\*.dll
```
dtivel, line 107: [.NET 8 simplified RID's](https://learn.microsoft.com/dotnet/core/compatibility/sdk/8.0/rid-graph) such that `win-x64` replaced `win10-x64`.

SdkTools.props
```diff
@@ -40,7 +40,7 @@
       <ScriptFilePath Condition=" '$(ScriptFilePath)' == '' ">$(RepositoryRootDirectory)\scripts\UpdateWintrust.ps1</ScriptFilePath>
     </PropertyGroup>
 
-    <Exec Command="$(PowerShellFilePath) -NonInteractive -ExecutionPolicy Unrestricted -Command &quot;&amp; { &amp;'$(ScriptFilePath)' '$(OutputPath)\tools\SDK\x64\wintrust.dll.ini' } &quot;" LogStandardErrorAsError="true" />
+    <Exec Command="$(PowerShellFilePath) -NonInteractive -ExecutionPolicy Unrestricted -Command &quot;&amp; { &amp;'$(ScriptFilePath)' '$(OutputPath)tools\SDK\x64\wintrust.dll.ini' } &quot;" LogStandardErrorAsError="true" />
```
dtivel, line 43: `$(OutputPath)` already has a trailing slash, so the output had an extra `\`.

`Adding NavSip.dll to wintrust.dll.ini - F:\git\sign\artifacts\bin\Sign.Cli\Debug\net8.0\win-x64\\tools\SDK\x64\wintrus
  t.dll.ini`

global.json
```diff
@@ -1,9 +1,9 @@
 {
   "tools": {
-    "dotnet": "8.0.100-alpha.1.23061.8",
+    "dotnet": "8.0.200",
```
dtivel, line 3: [The default `rollForward` policy is `latestPatch`](https://learn.microsoft.com/dotnet/core/tools/global-json#matching-rules), so this will grab latest `8.0.2xx.`

src/Sign.Core/Tools/VsixSignTool/XmlSignatureBuilder.cs
```diff
@@ -87,19 +87,25 @@ private Stream CanonicalizeElement(XmlElement element, out string canonicalizati
...
             //seems content to do an entire XmlDocument.
-            var transformer = new XmlDsigC14NTransform(false);
-            setCanonicalization?.Invoke(transformer.Algorithm);
 
-            var newDocument = new XmlDocument(_document.NameTable);
-            newDocument.LoadXml(element.OuterXml);
-            
-            transformer.LoadInput(newDocument);
+            var transformer = new XmlDsigC14NTransform(false);
+            string? algorithm = transformer.Algorithm;
```
dtivel, line 92: .NET 8 change `Transform.Algorithm`'s return type from `string` to `string?`, so this code raised a possible `null` error after updating to .NET 8.  A `null` or empty-length string is an invalid value for algorithm, so we should just fail fast if that's the case.

eng/Signing.props
```diff
@@ -16,7 +16,6 @@
     <FileSignInfo Include="AzureSign.Core.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="Newtonsoft.Json.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="NuGetKeyVaultSignTool.Core.dll" CertificateName="3PartySHA2" />
-    <FileSignInfo Include="OpenVsixSignTool.Core.dll" CertificateName="3PartySHA2" />
```
dtivel, line 19: This dependency was removed in https://github.com/dotnet/sign/commit/f08287270bf7010efb07bd34d6c6e9732ef0a3dd#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156L27

erdembayar (approved):

Issue #676, opened 2024-03-28
CI:  migrate to 1ES pipeline for internal CI builds


dtivel:
Steps:

- [x] Configure the public CI to use a PR-specific YAML file.  Currently, we have one YAML file for both internal and public builds.
- [x] Simplify the public CI YAML. 
- [X] Update the internal CI to use 1ES.
- [ ] Add baselines for SDL tooling.

PR #677, opened 2024-03-28
CI:  create separate public PR YAML

dtivel:
Progress on https://github.com/dotnet/sign/issues/676.

The PR copies /.vsts-ci.yml to ./.vsts-pr.yml then adds a comment to the top of each file indicating the AzDO pipeline URL.

### DIFF
--- a/.vsts-pr.yml
+++ b/.vsts-pr.yml
@@ -1,4 +1,4 @@
-# Pipeline: https://dnceng.visualstudio.com/internal/_build?definitionId=1190
+# Pipeline: https://dev.azure.com/dnceng-public/public/_build?definitionId=231
 
 variables:
   - name: _TeamName


.vsts-pr.yml
```diff
@@ -0,0 +1,150 @@
...
+    value: ${{ eq(variables['System.TeamProject'], 'internal') }}
+  - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+    - group: DotNet-Sign-SDLValidation-Params
+
+# CI and PR triggers
+trigger:
+  batch: true
+  branches:
+    include:
+    - main
```
jeffkl, line 20: In the past, I've been bitten by this if I am updating a servicing branch.  If this repo will use servicing branches, I recommend an entry for it like `rel/*` or `release-*`.  Obviously if this tool will never be servicing older versions, its not needed.

.vsts-pr.yml
```diff
@@ -0,0 +1,150 @@
...
+              mergeTestResults: true
+              searchFolder: $(System.DefaultWorkingDirectory)
+              testRunTitle: sign unit tests - $(Agent.JobName)
+            condition: succeededOrFailed()
+          - task: ComponentGovernanceComponentDetection@0
+            displayName: Component Governance scan
+            inputs:
+              ignoreDirectories: '$(Build.SourcesDirectory)/.packages,$(Build.SourcesDirectory)/artifacts/obj/Sign.Cli'
+
+- ${{ if eq(variables['System.TeamProject'], 'internal') }}:
```
jeffkl, line 126: Will this condition ever be true? I wonder if these make the pipeline more complex for no reason

.vsts-pr.yml
dtivel: Great feedback.  Currently, we don't have release branches.  We've only had prereleases.  We'll cross that bridge when we get there.

.vsts-pr.yml
dtivel: Yes, in the internal pipeline.

After this merges, I will create a new PR to simplify this file by removing official-only stuff.

jeffkl (approved):

Issue #689, opened 2024-05-31
CI broken


dtivel:
CI is failing with the following error.

```
Failed to request agent. Exception Image 1es-windows-2022-pt doesn't exist in pool NetCore1ESPool-Internal
```

See https://github.com/dotnet/aspnetcore/issues/55466

PR #690, opened 2024-05-31
CI:  update Arcade to fix CI error 

dtivel:
Resolve https://github.com/dotnet/sign/issues/689.

This change copies latest changes from https://github.com/dotnet/arcade/tree/release/8.0.

CC @clairernovotny, @erdembayar

### DIFF
--- a/eng/common/templates-official/job/publish-build-assets.yml
+++ b/eng/common/templates-official/job/publish-build-assets.yml
@@ -60,8 +60,8 @@ jobs:
       os: windows
     # If it's not devdiv, it's dnceng
     ${{ if ne(variables['System.TeamProject'], 'DevDiv') }}:
-      name: $(DncEngInternalBuildPool)
-      image: 1es-windows-2022-pt
+      name: NetCore1ESPool-Publishing-Internal
+      image: windows.vs2019.amd64
       os: windows
   steps:
   - ${{ if and(eq(parameters.runAsPublic, 'false'), ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
--- a/eng/common/templates-official/job/source-index-stage1.yml
+++ b/eng/common/templates-official/job/source-index-stage1.yml
@@ -1,6 +1,7 @@
 parameters:
   runAsPublic: false
-  sourceIndexPackageVersion: 1.0.1-20230228.2
+  sourceIndexUploadPackageVersion: 2.0.0-20240502.12
+  sourceIndexProcessBinlogPackageVersion: 1.0.1-20240129.2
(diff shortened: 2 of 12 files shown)


eng/common/templates-official/job/source-index-stage1.yml
```diff
@@ -62,7 +62,24 @@ jobs:
...
+    - task: AzureCLI@2
+      displayName: Get stage 1 auth token
+      inputs:
+        azureSubscription: 'SourceDotNet Stage1 Publish'
+        addSpnToEnvironment: true
+        scriptType: 'ps'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          echo "##vso[task.setvariable variable=ARM_CLIENT_ID]$env:servicePrincipalId"
+          echo "##vso[task.setvariable variable=ARM_ID_TOKEN]$env:idToken"
```
erdembayar, line 74: Not sure for Github, but Azure has this `;issecret=true` https://learn.microsoft.com/en-us/azure/devops/pipelines/process/set-secret-variables?view=azure-devops&tabs=yaml%2Cbash#set-secret-variable-in-a-script-with-logging-commands

Does it apply here?

`echo "##vso[task.setvariable variable=ARM_ID_TOKEN;issecret=true]$env:idToken"`

eng/common/templates-official/job/publish-build-assets.yml
```diff
@@ -60,8 +60,8 @@ jobs:
       os: windows
     # If it's not devdiv, it's dnceng
     ${{ if ne(variables['System.TeamProject'], 'DevDiv') }}:
-      name: $(DncEngInternalBuildPool)
-      image: 1es-windows-2022-pt
+      name: NetCore1ESPool-Publishing-Internal
+      image: windows.vs2019.amd64
```
erdembayar, line 64: Don't they have 2022 image?

eng/common/templates-official/job/source-index-stage1.yml
dtivel: It's a federated token, which isn't a secret in the traditional sense.  The whole idea behind federated credentials is to eliminate secret management.

dtivel: See https://github.com/dotnet/arcade/pull/14750/files#r1600647304

eng/common/templates-official/job/publish-build-assets.yml
dtivel: The source for this change is https://github.com/dotnet/arcade/pull/14663.

Is this important?  As long as CI builds keep working, it seems OK.  It has nothing to do with the .NET SDK version we use to build Sign CLI.  The Arcade team changed agent pools.  Maybe they're planning to update images within the pool.  A benefit of Arcade is that the Arcade team manages these details, so we don't have to.

erdembayar (approved):

PR #691, opened 2024-05-31
Remove SHA-1 thumbprint usage

javierdlg:
- Renames `--sha1` argument to a more generic `-cfp`/`--certificate-fingerprint` argument to allow the use of various RSA algorithms when searching for a matching certificate.
- Adds a new CLI argument `-cfpa`/`--certificate-fingerprint-algorithm` which allows specifying the certificate fingerprint algorithm used to find a matching certificate within the provided CSP
- Adds new tests for `-cfpa`
- Removes support for insecure algorithms (SHA 1 to 128)

### DIFF
--- a/src/Sign.Cli/CertificateStoreCommand.cs
+++ b/src/Sign.Cli/CertificateStoreCommand.cs
@@ -5,7 +5,6 @@
 using System.CommandLine;
 using System.CommandLine.Invocation;
 using System.CommandLine.IO;
-using System.Globalization;
 using System.Security.Cryptography;
 using System.Text;
 using Microsoft.Extensions.DependencyInjection;
@@ -87,10 +86,10 @@ namespace Sign.Cli
                 // SHA-1 Thumbprint is required in case the provided certificate container contains multiple certificates.
                 if (string.IsNullOrEmpty(sha1Thumbprint))
                 {
-                    context.Console.Error.WriteLine(
-                        FormatMessage(Resources.InvalidSha1ThumbprintValue, Sha1ThumbprintOption));
+                    context.Console.Error.WriteFormattedLine(
+                        Resources.InvalidSha1ThumbprintValue,
+                        Sha1ThumbprintOption);
                     context.ExitCode = ExitCode.NoInputsFound;
-
                     return;
                 }
 
@@ -114,10 +113,9 @@ namespace Sign.Cli
                 // Make sure this is rooted
                 if (!Path.IsPathRooted(baseDirectory.FullName))
                 {
-                    context.Console.Error.WriteLine(
-                        FormatMessage(
-                            Resources.InvalidBaseDirectoryValue,
-                            _codeCommand.BaseDirectoryOption));
+                    context.Console.Error.WriteFormattedLine(
+                        Resources.InvalidBaseDirectoryValue,
+                        _codeCommand.BaseDirectoryOption);
                     context.ExitCode = ExitCode.InvalidOptions;
                     return;
                 }
@@ -152,7 +150,6 @@ namespace Sign.Cli
                     {
                         context.Console.Error.WriteLine(Resources.InvalidFileValue);
                         context.ExitCode = ExitCode.InvalidOptions;
-
                         return;
                     }
 
@@ -207,10 +204,9 @@ namespace Sign.Cli
 
                 if (inputFiles.Any(file => !file.Exists))
                 {
-                    context.Console.Error.WriteLine(
-                        FormatMessage(
-                            Resources.SomeFilesDoNotExist,
-                            _codeCommand.BaseDirectoryOption));
+                    context.Console.Error.WriteFormattedLine(
+                        Resources.SomeFilesDoNotExist,
+                        _codeCommand.BaseDirectoryOption);
 
                     foreach (FileInfo file in inputFiles.Where(file => !file.Exists))
                     {
@@ -247,15 +243,6 @@ namespace Sign.Cli
             }
 
             return Path.Combine(baseDirectory.FullName, file);
-        }
-
-        private static string FormatMessage(string format, params IdentifierSymbol[] symbols)
-        {
-            string[] formattedSymbols = symbols
-                .Select(symbol => $"--{symbol.Name}")
-                .ToArray();
-
-            return string.Format(CultureInfo.CurrentCulture, format, formattedSymbols);
         }
     }
 }
--- a/src/Sign.Cli/CertificateStoreCommand.cs
+++ b/src/Sign.Cli/CertificateStoreCommand.cs
@@ -92,34 +92,11 @@ namespace Sign.Cli
                 if (string.IsNullOrEmpty(certificateFingerprint))
                 {
                     context.Console.Error.WriteFormattedLine(
-                        Resources.InvalidCertificateThumbprintValue,
-                        CertificateThumbprintOption);
+                        Resources.InvalidCertificateFingerprintValue,
+                        CertificateFingerprintOption);
                     context.ExitCode = ExitCode.NoInputsFound;
 
                     return;
-                }
-
-                HashAlgorithmName certificateFingerprintAlgorithm = HashAlgorithmName.SHA256;
-
-                // CertificateFingerprintAlgorithm defaults to SHA-256, but one could pass in an empty argument through PowerShell scripts.
-                if (unparsedCertificateFingerprintAlgorithm != null)
-                {
-                    // Remove "-" from the unparsed entry to match with HashAlgorithmName.
-                    unparsedCertificateFingerprintAlgorithm = string.Join(string.Empty, unparsedCertificateFingerprintAlgorithm.Split('-'));
-
-                    // We only accept SHA256, SHA384, and SHA512 as valid fingerprint algorithms.
-                    if (!Enum.TryParse<HashAlgorithmName>(unparsedCertificateFingerprintAlgorithm, ignoreCase: true, out certificateFingerprintAlgorithm)
-                        && (certificateFingerprintAlgorithm != HashAlgorithmName.SHA256
-                            || certificateFingerprintAlgorithm != HashAlgorithmName.SHA384
-                            || certificateFingerprintAlgorithm != HashAlgorithmName.SHA512))
-                    {
-                        context.Console.Error.WriteFormattedLine(
-                            Resources.InvalidCertificateFingerprintAlgorithmValue,
-                            CertificateFingerprintAlgorithmOption);
-                        context.ExitCode = ExitCode.NoInputsFound;
-
-                        return;
-                    }
                 }
 
                 // CSP requires a private key container to function.
--- a/src/Sign.Cli/CertificateStoreCommand.cs
+++ b/src/Sign.Cli/CertificateStoreCommand.cs
@@ -78,7 +78,6 @@ namespace Sign.Cli
                 bool useMachineKeyContainer = context.ParseResult.GetValueForOption(UseMachineKeyContainerOption);
 
                 string? fileArgument = context.ParseResult.GetValueForArgument(FileArgument);
-                HashAlgorithmName certificateFingerprintAlgorithm = HashAlgorithmName.SHA256;
 
                 if (string.IsNullOrEmpty(fileArgument))
                 {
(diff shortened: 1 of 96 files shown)


src/Sign.Cli/CertificateStoreResources.resx
```diff
@@ -121,6 +121,14 @@
     <value>PFX, P7B, or CER file containing a certificate and potentially a private key.</value>
     <comment>{Locked="PFX", "P7B", "CER"} are file extensions.</comment>
   </data>
+  <data name="CertificateFingerprintAlgorithmOptionDescription" xml:space="preserve">
+    <value>SHA fingerprint algorithm. Only accepts SHA-256, SHA-384, or SHA-512</value>
+    <comment>{Locked="SHA-256", "SHA-384", "SHA-512"} are cryptographic algorithms.</comment>
```
dtivel, line 126:
```suggestion
    <value>Certificate fingerprint algorithm. Allowed values are 'sha256', 'sha384', and 'sha512'.</value>
    <comment>{Locked="sha256", "sha384", "sha512"} are cryptographic hash algorithms.</comment>
```

To be consistent with an existing option, see https://github.com/dotnet/sign/blob/27cf6bdb045d6d4efe04f61895b9b5c5a67b0fdf/src/Sign.Cli/Resources.resx#L138-L141.

src/Sign.Core/Resources.resx
```diff
@@ -132,7 +132,7 @@
     <value>Unable to find a matching certificate.</value>
   </data>
   <data name="CertificateNotFoundInFile" xml:space="preserve">
-    <value>Unable to find a certificate with a matching SHA-1 thumbprint in {0}.</value>
+    <value>Unable to find a certificate with a matching {0} fingerprint in {1}.</value>
     <comment>{NumberedPlaceholder="{0}"} is a file path.</comment>
```
dtivel, line 136:
```suggestion
    <comment>{NumberedPlaceholder="{0}"} is a certificate fingerprint.  {NumberedPlaceholder="{1}"} is a file path.</comment>
```

src/Sign.Core/CertificateServices/CertificateStoreService.cs
```diff
@@ -134,8 +137,8 @@ private Task<X509Certificate2> GetStoreCertificateAsync()
             // Search User or Machine certificate stores.
```
dtivel, line 137: I can't comment directly on (new) line 134 because it wasn't changed; however, [you updated the `CertificateNotFoundInFile` resource string to take 2 string replacements](https://github.com/dotnet/sign/pull/691/files#diff-f0243fdd6fbbfb540457054900a66fd592aca55b3410d47e651313b479207b43R135), but you're still passing one argument here.

src/Sign.Core/CertificateServices/CertificateStoreService.cs
```diff
@@ -134,8 +137,8 @@ private Task<X509Certificate2> GetStoreCertificateAsync()
...
             if (!string.IsNullOrEmpty(_privateKeyContainer)
                 && _isPrivateMachineKeyContainer
-                    ? TryFindCertificate(StoreLocation.LocalMachine, _sha1Thumbprint!, out X509Certificate2? certificate)
-                    : TryFindCertificate(StoreLocation.CurrentUser, _sha1Thumbprint!, out certificate))
+                    ? TryFindCertificate(StoreLocation.LocalMachine, _certificateFingerprint!, out X509Certificate2? certificate)
+                    : TryFindCertificate(StoreLocation.CurrentUser, _certificateFingerprint!, out certificate))
```
dtivel, line 141: or here

src/Sign.Core/CertificateServices/CertificateStoreServiceProvider.cs
```diff
@@ -88,7 +94,15 @@ private CertificateStoreService GetService(IServiceProvider serviceProvider)
...
-                _certificateStoreService = new CertificateStoreService(serviceProvider, _sha1Thumbprint, _cryptoServiceProvider, _privateKeyContainer, _certificateFilePath, _certificateFilePassword, _isMachineKeyContainer);
+                _certificateStoreService = new CertificateStoreService(
+                                            serviceProvider,
+                                            _certificateFingerprint,
+                                            _certificateFingerprintAlgorithm,
+                                            _cryptoServiceProvider,
+                                            _privateKeyContainer,
+                                            _certificateFilePath,
+                                            _certificateFilePassword,
+                                            _isMachineKeyContainer);
```
dtivel, line 105:
```suggestion
                _certificateStoreService = new CertificateStoreService(
                    serviceProvider,
                    _certificateFingerprint,
                    _certificateFingerprintAlgorithm,
                    _cryptoServiceProvider,
                    _privateKeyContainer,
                    _certificateFilePath,
                    _certificateFilePassword,
                    _isMachineKeyContainer);
```

src/Sign.Core/Resources.resx
```diff
@@ -190,8 +190,8 @@
   <data name="FetchingCertificate" xml:space="preserve">
     <value>Fetching certificate from Azure Key Vault.</value>
   </data>
-  <data name="InvalidSha1ThumbrpintValue" xml:space="preserve">
-    <value>Invalid value for {0}. The value must be the SHA-1 thumbprint of a certificate in a Windows Certificate Store.</value>
+  <data name="InvalidCertificateFingerprintValue" xml:space="preserve">
+    <value>Invalid value for {0}. The value must be a {1} fingerprint of a certificate in a Windows Certificate Store.</value>
```
dtivel, line 194: You updated this resource string to have 2 string replacements instead of 1, but the call site still passes 1 argument.

src/Sign.Core/CertificateServices/CertificateStoreService.cs
```diff
@@ -147,17 +150,17 @@ private Task<X509Certificate2> GetStoreCertificateAsync()
...
 
-        private static bool TryFindCertificate(StoreLocation storeLocation, string sha1Fingerprint, [NotNullWhen(true)] out X509Certificate2? certificate)
+        private bool TryFindCertificate(StoreLocation storeLocation, string certificateFingerprint, [NotNullWhen(true)] out X509Certificate2? certificate)
         {
             // Check machine certificate store.
             using (var store = new X509Store(StoreName.My, storeLocation))
             {
                 store.Open(OpenFlags.OpenExistingOnly | OpenFlags.ReadOnly);
-                var certificates = store.Certificates.Find(X509FindType.FindByThumbprint, sha1Fingerprint, validOnly: false);
+                var certificates = store.Certificates.Find(X509FindType.FindByThumbprint, certificateFingerprint, validOnly: false);
```
dtivel, line 159: Have you verified that this works?  You should have tests for it.

Looking at [.NET code](https://github.com/dotnet/runtime/blob/736ca4bede96c2cc98a13da8f8dfd6bb8688f15d/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/X509Certificates/ManagedCertificateFinder.cs#L40-L63), it looks like `FindByThumbprint` only searches by SHA-1 fingerprint, so passing anything else would just fail.

src/Sign.Cli/CertificateStoreCommand.cs
```diff
@@ -84,16 +87,38 @@ internal CertificateStoreCommand(CodeCommand codeCommand, IServiceProviderFactor
...
+                HashAlgorithmName certificateFingerprintAlgorithm = HashAlgorithmName.SHA256;
+
+                // CertificateFingerprintAlgorithm defaults to SHA-256, but one could pass in an empty argument through PowerShell scripts.
+                if (unparsedCertificateFingerprintAlgorithm != null)
+                {
+                    // Remove "-" from the unparsed entry to match with HashAlgorithmName.
+                    unparsedCertificateFingerprintAlgorithm = string.Join(string.Empty, unparsedCertificateFingerprintAlgorithm.Split('-'));
+
+                    // We only accept SHA256, SHA384, and SHA512 as valid fingerprint algorithms.
+                    if (!Enum.TryParse<HashAlgorithmName>(unparsedCertificateFingerprintAlgorithm, ignoreCase: true, out certificateFingerprintAlgorithm)
```
dtivel, line 109: `System.Security.Cryptography.HashAlgorithmName` is a struct not an enum, so `Enum.TryParse<HashAlgorithmName>(...)` will throw a `System.ArgumentException` exception.  This code doesn't work.

clairernovotny (changes requested):

src/Sign.Cli/CertificateStoreCommand.cs
```diff
@@ -84,16 +87,38 @@ internal CertificateStoreCommand(CodeCommand codeCommand, IServiceProviderFactor
...
 
+                HashAlgorithmName certificateFingerprintAlgorithm = HashAlgorithmName.SHA256;
+
+                // CertificateFingerprintAlgorithm defaults to SHA-256, but one could pass in an empty argument through PowerShell scripts.
+                if (unparsedCertificateFingerprintAlgorithm != null)
+                {
+                    // Remove "-" from the unparsed entry to match with HashAlgorithmName.
+                    unparsedCertificateFingerprintAlgorithm = string.Join(string.Empty, unparsedCertificateFingerprintAlgorithm.Split('-'));
+
+                    // We only accept SHA256, SHA384, and SHA512 as valid fingerprint algorithms.
```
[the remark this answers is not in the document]
javierdlg: Guidance is to not use SHA-1 for any features. So we have to forcibly not allow them as options:
https://liquid.microsoft.com/Web/Object/Read/MS.Security/Requirements/Microsoft.Security.Cryptography.10021#Zguidance

I'll be updating our documentation with how to access SHA256+ fingerprints. But maybe we can point them to guidance if we see them try to use SHA-1 to keep this user-friendly for newcomers?

src/Sign.Cli/CertificateStoreCommand.cs
```diff
@@ -84,16 +87,38 @@ internal CertificateStoreCommand(CodeCommand codeCommand, IServiceProviderFactor
...
+                        && (certificateFingerprintAlgorithm != HashAlgorithmName.SHA256
+                            || certificateFingerprintAlgorithm != HashAlgorithmName.SHA384
+                            || certificateFingerprintAlgorithm != HashAlgorithmName.SHA512))
+                    {
+                        context.Console.Error.WriteLine(
+                            FormatMessage(Resources.InvalidCertificateFingerprintAlgorithmValue, CertificateFingerprintAlgorithmOption));
+                        context.ExitCode = ExitCode.NoInputsFound;
+
+                        return;
+                    }
```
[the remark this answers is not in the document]
javierdlg: This is great, it'll simplify parsing as well

src/Sign.Cli/CertificateStoreCommand.cs
```diff
@@ -20,7 +20,8 @@ internal sealed class CertificateStoreCommand : Command
     {
         private readonly CodeCommand _codeCommand;
 
-        internal Option<string> Sha1ThumbprintOption { get; } = new(new[] { "-s", "--sha1" }, CertificateStoreResources.Sha1ThumbprintOptionDescription);
+        internal Option<string> CertificateFingerprintOption { get; } = new(new[] { "-cfp", "--certificate-fingerprint" }, CertificateStoreResources.CertificateFingerprintOptionDescription);
```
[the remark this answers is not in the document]
javierdlg: Not yet released luckily

src/Sign.Cli/CertificateStoreCommand.cs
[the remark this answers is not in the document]
javierdlg: There's no good documentation website on this. So I'll add it to our VS docs and then open a PR here to point them there on how to get a SHA256+ hash from their certificate files.

src/Sign.Core/CertificateServices/CertificateStoreService.cs
```diff
@@ -28,25 +30,33 @@ internal sealed class CertificateStoreService : ISignatureAlgorithmProvider, ICe
...
             _certificatePath = certificatePath;
             _certificatePassword = certificatePassword;
 
+            _hasher = _certificateFingerprintAlgorithm.Name switch
+            {
+                "SHA256" => SHA256.Create(),
+                "SHA384" => SHA384.Create(),
+                "SHA512" => SHA512.Create(),
+
+                _ => throw new NotSupportedException(Resources.VSIXSignToolUnsupportedAlgorithm)
```
dtivel, line 57: This will go away with my other suggested change in this file.  However, I wanted to comment on this.

I see the `CertificateStoreService` as being general purpose.  Yes, VSIX signing uses it, but anything should be able to use it.  The name of the resource string implies a VSIX restriction is being respected here.  (The message itself is not VSIX-specific.)

Ideally, at a global level, we should be able to allow or disallow specific algorithms, but enable the broadest set of options by default, while individual signature formatters (e.g.:  NuGet, VSIX, etc.) should be able to indicate they do or do not support specific algorithms.

This location is completely independent of signature format, but the resource string indicates otherwise.

dtivel (changes requested):
Looking much better.  Thanks for the updates.

dtivel:
@javierdlg, be sure to rebase on latest main.

src/Sign.Core/CertificateServices/CertificateStoreService.cs
```diff
@@ -147,17 +158,17 @@ private Task<X509Certificate2> GetStoreCertificateAsync()
...
             {
                 store.Open(OpenFlags.OpenExistingOnly | OpenFlags.ReadOnly);
-                var certificates = store.Certificates.Find(X509FindType.FindByThumbprint, sha1Fingerprint, validOnly: false);
 
-                foreach (var cert in certificates)
+                foreach (var cert in store.Certificates)
                 {
-                    if (string.Equals(cert.Thumbprint, sha1Fingerprint, StringComparison.InvariantCultureIgnoreCase))
+                    // X509Certificate2 only exposes a SHA-1 thumbprint. To circumvent this, generate a hash that matches the specified algorithm name from the raw certificate bytes.
+                    if (string.Equals(CalculateHashFromRawData(cert.GetRawCertData()), certificateFingerprint, StringComparison.InvariantCultureIgnoreCase))
```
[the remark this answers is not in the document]
javierdlg: Looks like when I F12 VS erroneously took me to the wrong method which returns the SHA-1 thumbprint, leading to me calculating this hash manually... I can see now that passing in an HashAlgorithmName does give you that computed hash

src/Sign.Core/Resources.resx
```diff
@@ -190,9 +190,9 @@
   <data name="FetchingCertificate" xml:space="preserve">
     <value>Fetching certificate from Azure Key Vault.</value>
   </data>
-  <data name="InvalidSha1ThumbrpintValue" xml:space="preserve">
-    <value>Invalid value for {0}. The value must be the SHA-1 thumbprint of a certificate in a Windows Certificate Store.</value>
-    <comment>{NumberedPlaceholder="{0}"} is a a SHA-1 thumbprint and shouldn't be translated.</comment>
+  <data name="InvalidCertificateFingerprintValue" xml:space="preserve">
```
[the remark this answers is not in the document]
dtivel: Make sure you run eng\common\CIBuild.cmd after removing this resource string (through the designer).

dtivel (approved):

clairernovotny (approved):

PR #693, opened 2024-06-03
Added extension method that replaces FormatMessage in the two command classes.

dlemstra:
This avoids the code duplication of the FormatMessage method. This is a first step in removing some of the duplication of the `AzureKeyVaultCommand` and `CertificateStoreCommand` that is also needed by a new `TrustedSigningCommand`.

### DIFF
--- a/src/Sign.Cli/StandardStreamWriterExtensions.cs
+++ b/src/Sign.Cli/StandardStreamWriterExtensions.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+using System.CommandLine;
+using System.CommandLine.IO;
+using System.Globalization;
+
+namespace Sign.Cli
+{
+    internal static class StandardStreamWriterExtensions
+    {
+        internal static void WriteFormattedLine(this IStandardStreamWriter writer, string format, params IdentifierSymbol[] symbols)
+        {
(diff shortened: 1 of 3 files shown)


src/Sign.Cli/StandardStreamWriterExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    internal static class StandardStreamWriterExtensions
+    {
+        internal static void WriteFormattedLine(this IStandardStreamWriter writer, string format, params IdentifierSymbol[] symbols)
+        {
+            string[] formattedSymbols = symbols
+                .Select(symbol => $"--{symbol.Name}")
+                .ToArray();
+
+            writer.WriteLine(string.Format(CultureInfo.CurrentCulture, format, formattedSymbols));
```
clairernovotny, line 19: Shouldn't this be `InvariantCulture` as these are parameters and not user input?

clairernovotny: Please, in this case

dlemstra: I have rebased the PR and update the commit to include this change.

clairernovotny (approved):

src/Sign.Cli/StandardStreamWriterExtensions.cs
clairernovotny: Looks like it's still showing CurrentCulture?

clairernovotny (approved):

PR #700, opened 2024-06-07
Renamed SignatureProvider to DataFormatSigner so SignatureProvider can be used by the new SignatureProvider libraries.

dlemstra:
The name SignatureProvider is now use by the classes that are DataFormatSigners and we want this name to be used in the SignatureProvider libraries. Renaming all these classes and interfaces will make it possible to use this name.


clairernovotny (approved):

dtivel:
@clairernovotny, I was still reviewing this when it merged.  :(

@dlemstra, can you please rename the folder https://github.com/dotnet/sign/tree/main/src/Sign.Core/SignatureProviders to DataFormatSigners?

clairernovotny:
Eek, sorry!

dlemstra:
Would you be okay with me renaming that folder in a PR where I move the `NuGetSigner` to a separate library? Or would you like me to create a separate PR for this?

dlemstra:
I noticed that there were some issues when moving the NuGetSigner to a separate library. One of the is an incorrect ILogger injection. To address this and your comment I created a PR to fix this. For the other issue I will create a separate PR.

PR #708, opened 2024-06-13
Housekeeping

dlemstra:
This PR does some housekeeping to fix the following "issues":

- Correct the ILogger injections and fields types.
- Rename the folder that was mentioned as a review comment in #700.
- Use a consistent constructor notation through the whole project.


dlemstra:
Just realized that we could use `$(MSBuildThisFileDirectory)` for `RepositoryRootDirectory` inside the `Directory.Build.props` file. Would it be okay to also add that to this PR?

dtivel:
> Just realized that we could use `$(MSBuildThisFileDirectory)` for the value of `RepositoryRootDirectory` inside the `Directory.Build.props` file. Would it be okay to also add that to this PR?

Sure, sounds good.  Thanks, @dlemstra.

dlemstra:
I have updated the PR and also included some changes to the `.editorconfig` because I kept getting tabs when pasting something in the `.csproj` files.

dtivel (approved):

PR #713, opened 2024-06-13
Moved the shared logic of the AzureKeyVaultCommand and CertificateStoreCommand to a new Handle method inside the CodeCommand

dlemstra:
This PR moves the shared logic of the AzureKeyVaultCommand and CertificateStoreCommand to a new Handle method inside the CodeCommand. And this also introduces a new interface called ISignatureProvider that can be used for a new TrustedSigningServiceProvider class.

### DIFF
--- a/src/Sign.Cli/CodeCommand.cs
+++ b/src/Sign.Cli/CodeCommand.cs
@@ -3,10 +3,17 @@
 // See the LICENSE.txt file in the project root for more information.
 
 using System.CommandLine;
+using System.CommandLine.Invocation;
+using System.CommandLine.IO;
 using System.CommandLine.Parsing;
 using System.Globalization;
 using System.Security.Cryptography;
+using System.Text;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.FileSystemGlobbing;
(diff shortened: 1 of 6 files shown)


src/Sign.Cli/CodeCommand.cs
```diff
@@ -54,6 +61,145 @@ internal CodeCommand()
             AddGlobalOption(VerbosityOption);
         }
 
+        internal async Task Handle(InvocationContext context, IServiceProviderFactory serviceProviderFactory, ISignatureProvider signatureProvider, string fileArgument)
```
dtivel, line 64:
```suggestion
        internal async Task HandleAsync(InvocationContext context, IServiceProviderFactory serviceProviderFactory, ISignatureProvider signatureProvider, string fileArgument)
```

dtivel (changes requested):

src/Sign.Cli/CodeCommand.cs
dlemstra: Thanks! I have updated my commit and rebased it.

dtivel (approved):

PR #715, opened 2024-06-14
Use ConcurrentBag instead of a List because the .Add method of List is not thread safe.

dlemstra:
This PR fixes a mistake that I made in my earlier pull request. I did not take in account that List.Add is not thread safe. Using a ConcurrentBag instead fixes this issue.

### DIFF
--- a/test/Sign.SignatureProviders.KeyVault.Test/KeyVaultServiceProviderTests.cs
+++ b/test/Sign.SignatureProviders.KeyVault.Test/KeyVaultServiceProviderTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE.txt file in the project root for more information.
 
+using System.Collections.Concurrent;
 using Azure.Core;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


test/Sign.SignatureProviders.KeyVault.Test/KeyVaultServiceProviderTests.cs
```diff
@@ -77,14 +78,14 @@ public void GetSignatureAlgorithmProvider_ReturnsSameInstance()
         {
             KeyVaultServiceProvider provider = new(tokenCredential, KeyVaultUrl, CertificateName);
 
-            List<ISignatureAlgorithmProvider> signatureAlgorithmProviders = [];
+            ConcurrentBag<ISignatureAlgorithmProvider> signatureAlgorithmProviders = [];
```
dtivel, line 81: These tests would be simpler without parallelism.  I think parallelism adds unnecessary complexity.  (Hence, the bug.)

See https://github.com/dotnet/sign/blob/dc00170e656abc2c5f17331a22196de096a2bb53/test/Sign.Core.Test/FileSystem/FileInfoComparerTests.cs#L12-L18.

dlemstra: I added parallelism to also test the lock inside those methods. Otherwise I would do it as you suggested.

dtivel (approved):

Issue #683, opened 2024-04-29
Support for Azure Trusted Signing


EuanKerr:
With [Azure Trusted Signing](https://learn.microsoft.com/en-us/azure/trusted-signing/) (formerly Azure Code Signing) now in public preview it would be handy to be able to sign files/clickonce using the Trusted Signing service.

Some more links:
- https://www.powershellgallery.com/packages/TrustedSigning

dlemstra:
Would love to see support for this. The @ImageMagick project has a Trusted Signing certificate and we are using this to sign our installers and executables. But there is no support for that in this tool yet. And I would like to also sign my ImageMagick NuGet packages using that certificate in a GitHub pipeline.

blowdart:
Throwing my weight behind this as I have a new package I want to test with and I'm trying to keep everything in GitHub actions ad an added bonus.

ROGG437063:
I am also supporting this request, our developers need to migrate to trusted signing, but without support for clickonce in any of the other trusted signing tools, that is not possible.

ROGG437063:
This integration should probably use the SDK https://www.nuget.org/packages/Microsoft.Trusted.Signing.Client
Do take care that the SDK linked in trusted signing docs is deprecated

ianjmcm:
@ROGG437063 sorry for the confusion on the docs regarding the SDK. We have kept the current version from our preview unlisted while we prep a new release of the SDK. This SDK will become publicly listed in the very near future. https://www.nuget.org/packages/Azure.CodeSigning.Sdk 
cc: @Jaxelr

Jaxelr:
@ROGG437063  Indeed, we expect the SDK to be publicly available once we sort through a few issues on the PR, but you can track that work here: https://github.com/Azure/azure-sdk-for-net/pull/41191

ROGG437063:
@Jaxelr Sounds good, any estimate of when that will be merged? It would also still need integration with sign cli for vsto manifest signing, right?

demiliani:
Must to have. Please push 🙌🏻

dtivel:
[This prerelease version](https://www.nuget.org/packages/sign/0.9.1-beta.24325.5) has Trusted Signing support.  Thanks to @dlemstra for adding it!  👏

mjcheetham:
[...]

Issue #561, opened 2023-01-16
Signing VSIX may fail with `Azure.Identity.CredentialUnavailableException` when Visual Studio is also installed


dtivel:
## Repro steps

1. Download InstallerProjects.vsix from https://marketplace.visualstudio.com/items?itemName=VisualStudioClient.MicrosoftVisualStudio2017InstallerProjects.
2. Execute

   ```text
   sign code azure-key-vault -kvu **** -kvc MyCertificate -kvm -d Description -u http://description.test -t http://timestamp.digicert.com -b E:\testpass .\InstallerProjects.vsix
   ```

## Result

```PS
info: Sign.Core.Signer[0]
      Submitting E:\testpass\InstallerProjects.vsix for signing.
info: Sign.Core.Signer[0]
      SignAsync called for E:\testpass\InstallerProjects.vsix. Using C:\Users\dtivel\AppData\Local\Temp\qgueuuhu.q23\cyrpatbb.vsix locally.
info: Sign.Core.IDirectoryService[0]
      Extracting container C:\Users\dtivel\AppData\Local\Temp\qgueuuhu.q23\cyrpatbb.vsix to C:\Users\dtivel\AppData\Local\Temp\i5gmg5o2.2cc.
info: Sign.Core.ISignatureProvider[0]
      Signing SignTool job with 21 files.
info: Sign.Core.ISignatureProvider[0]
      Signing SignTool job with 4 files.
info: Sign.Core.ISignatureProvider[0]
      Signing C:\Users\dtivel\AppData\Local\Temp\i5gmg5o2.2cc\bin\dpplg.dll.
info: Sign.Core.ISignatureProvider[0]
      Signing C:\Users\dtivel\AppData\Local\Temp\i5gmg5o2.2cc\bin\dpedt.dll.
info: Sign.Core.ISignatureProvider[0]
      Signing C:\Users\dtivel\AppData\Local\Temp\i5gmg5o2.2cc\bin\DPCA.dll.
info: Sign.Core.ISignatureProvider[0]
      Signing C:\Users\dtivel\AppData\Local\Temp\i5gmg5o2.2cc\$PublicAssemblies\Microsoft.VisualStudio.DeployWizard.dll.
info: Sign.Core.ISignatureProvider[0]
      Signing C:\Users\dtivel\AppData\Local\Temp\i5gmg5o2.2cc\VsDeploy\dpdplidl.tlb.
info: Sign.Core.ISignatureProvider[0]
[...]

PR #716, opened 2024-06-14
Added support for Trusted Signing

dlemstra:
This PR adds support for Trusted Signing that was requested in #683 with the help of the Azure.CodeSigning.Sdk library.


src/Sign.Cli/TrustedSigningCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Sign.Cli
+{
+    internal sealed class TrustedSigningCommand : Command
+    {
+        internal Option<Uri> EndpointOption { get; } = new(["-tse", "--trusted-signing-endpoint"], TrustedSigningResources.EndpointOptionDescription);
+        internal Option<string> AccountOption { get; } = new(["-tsa", "--trusted-signing-account"], TrustedSigningResources.AccountOptionDescription);
+        internal Option<string> CertificateProfileOption { get; } = new(["-tsc", "--trusted-signing-certificate-profile"], TrustedSigningResources.CertificateProfileOptionDescription);
+        internal Option<bool> ManagedIdentityOption { get; } = new(["-tsm", "--trusted-signing-managed-identity"], getDefaultValue: () => false, TrustedSigningResources.ManagedIdentityOptionDescription);
+        internal Option<string?> TenantIdOption { get; } = new(["-tst", "--trusted-signing-tenant-id"], TrustedSigningResources.TenantIdOptionDescription);
```
clairernovotny, line 21: Do we want to consider a common base of common parameters across providers? In this case, Key Vault and Trusted Signing use these same values. There's clearly a balance between keeping providers separate vs duplicating code/functionality/parameters.

Another thought is whether Auth should be a separate set of parameters/provider? Entra would be one in-box option (using Azure.Identity) that both KV and TS providers use. Certificate store doesn't need auth so that's easy. Potentially another provider could use Entra auth or provide another auth mechanism that's specific to it. 

Should we factor out auth into its own thing?

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+            _mutex.Dispose();
+            _publicKey?.Dispose();
+            GC.SuppressFinalize(this);
+        }
+
+        public async Task<X509Certificate2> GetCertificateAsync(CancellationToken cancellationToken)
+        {
+            if (_publicKey is not null)
+            {
+                return new X509Certificate2(_publicKey);
```
clairernovotny, line 61: Can this be cached/reused so that multiple requests don't new up additional certificate objects

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+
+                    _logger.LogTrace(Resources.FetchedCertificate, stopwatch.Elapsed.TotalMilliseconds);
+                }
+            }
+            finally
+            {
+                _mutex.Release();
+            }
+
+            return new X509Certificate2(_publicKey);
```
clairernovotny, line 90: Same as prior -- lets try reuse the cert object instance

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        public async Task<RSA> GetRsaAsync(CancellationToken cancellationToken)
+        {
+            X509Certificate2 publicKey = await GetCertificateAsync(cancellationToken);
+            return new RSATrustedSigning(_client, _accountName, _certificateProfileName, publicKey);
```
clairernovotny, line 96: Can this be reused as well?

dlemstra:
The build is failing because if this error:

```
src\Sign.SignatureProviders.TrustedSigning\Sign.SignatureProviders.TrustedSigning.csproj(0,0): error NU1101: (NETCORE_ENGINEERING_TELEMETRY=Restore) Unable to find package Azure.CodeSigning.Sdk. No packages exist with this id in source(s): dotnet-8, dotnet-eng, dotnet-libraries, dotnet-public, dotnet-tools
````

I have this package in my local cache and that is why "it works" on my machine. Which source should I add to make this work?

File: src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs (hunk 2 above)
Comment (dlemstra) [reply src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs:61]: When I do that I get the following error:

```
fail: Sign.Core.ISigner[0]
      The chain context handle is invalid. (Parameter 'certificate')
      System.ArgumentException: The chain context handle is invalid. (Parameter 'certificate')
         at System.Security.Cryptography.X509Certificates.X509Chain.Build(X509Certificate2 certificate, Boolean throwOnException)
         at AzureSign.Core.AuthenticodeKeyVaultSigner..ctor(AsymmetricAlgorithm signingAlgorithm, X509Certificate2 signingCertificate, HashAlgorithmName fileDigestAlgorithm, TimeStampConfiguration timeStampConfiguration, X509Certificate2Collection additionalCertificates)
         at Sign.Core.AzureSignToolSigner.SignAsync(IEnumerable`1 files, SignOptions options) in D:\PullRequests\dotnet\sign\src\Sign.Core\DataFormatSigners\AzureSignToolSigner.cs:line 96
         at Sign.Core.AggregatingSigner.SignAsync(IEnumerable`1 files, SignOptions options) in D:\PullRequests\dotnet\sign\src\Sign.Core\DataFormatSigners\AggregatingSigner.cs:line 204
         at Sign.Core.AggregatingSigner.SignAsync(IEnumerable`1 files, SignOptions options) in D:\PullRequests\dotnet\sign\src\Sign.Core\DataFormatSigners\AggregatingSigner.cs:line 92
         at Sign.Core.Signer.<>c__DisplayClass3_0.<<SignAsync>b__0>d.MoveNext() in D:\PullRequests\dotnet\sign\src\Sign.Core\Signer.cs:line 155
      --- End of stack trace from previous location ---
         at System.Threading.Tasks.Parallel.<>c__53`1.<<ForEachAsync>b__53_0>d.MoveNext()
      --- End of stack trace from previous location ---
         at Sign.Core.Signer.SignAsync(IReadOnlyList`1 inputFiles, String outputFile, FileInfo fileList, DirectoryInfo baseDirectory, String applicationName, String publisherName, String description, Uri descriptionUrl, Uri timestampUrl, Int32 maxConcurrency, HashAlgorithmName fileHashAlgorithm, HashAlgorithmName timestampHashAlgorithm) in D:\PullRequests\dotnet\sign\src\Sign.Core\Signer.cs:line 84
```

That is happening because we are doing `using (X509Certificate2 certificate = await _certificateProvider.GetCertificateAsync())` inside `AzureSignToolSigner.SignAsync`.

File: src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs (hunk 4 above)
Comment (dlemstra) [reply src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs:96]: The object will be `Disposed` so we cannot reuse it.

File: src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs (hunk 3 above)
Comment (dlemstra) [reply src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs:90]: See my comment at the other spot

File: src/Sign.Cli/TrustedSigningCommand.cs (hunk 1 above)
Comment (dlemstra) [reply src/Sign.Cli/TrustedSigningCommand.cs:21]: We could move this out and share it between these two. But that would mean we would need to pass in the `Option` objects to that method to get proper error reporting?

**Comment by Jaxelr** [comment]
Isn't the package publicly available in nuget? https://www.nuget.org/packages/Azure.CodeSigning.Sdk

File: src/Sign.Cli/TrustedSigningCommand.cs [hunk 5]
```diff
@@ -0,0 +1,100 @@
...
+        internal Option<string?> ClientIdOption { get; } = new(["-tsi", "--trusted-signing-client-id"], TrustedSigningResources.ClientIdOptionDescription);
+        internal Option<string?> ClientSecretOption { get; } = new(["-tss", "--trusted-signing-client-secret"], TrustedSigningResources.ClientSecretOptionDescription);
```
Jaxelr, line 23: This is probably not the scope of this PR, but are we taking into account what's our plan for future experiences given we want to move away from secrets and be able to provide an experience for example for federated credentials ?

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+            await _mutex.WaitAsync();
+            try
+            {
+                if (_publicKey is null)
+                {
+                    Stopwatch stopwatch = Stopwatch.StartNew();
+
+                    _logger.LogTrace(Resources.FetchingCertificate);
+
+                    CertificateProfileSignOperation operation = await _client.StartSignAsync(_accountName, _certificateProfileName, _emptyRequest, cancellationToken: cancellationToken);
```
Jaxelr, line 73: I think I mentioned this offline, but we should be looking to provide an updated sdk which would include the get certificate chain operation, that way we could find the leaf cert being used for signing prior to submitting and avoid the incurred cost of a dummy get operation.

src/Sign.Cli/TrustedSigningCommand.cs
clairernovotny: Why would we need to pass option objects around? I was thinking of a new auth entity/abstraction that handles these? Ultimately both TS and KV use Azure.Identity with either a DefaultAzureCredential or TokenCredential, so I imagine there's a good way to do this>

dlemstra:
> @dlemstra Isn't the package publicly available in nuget? https://www.nuget.org/packages/Azure.CodeSigning.Sdk

The nuget.org feed is not included in the `NuGet.Config` file of this project. Wondering if I can add it?

src/Sign.Cli/TrustedSigningCommand.cs
dlemstra: Maybe something like this:

```C#
internal sealed class AzureAutenticationOptions
{
    internal Option<bool> ManagedIdentityOption { get; } = new(["-ami", "--azure-managed-identity"], getDefaultValue: () => false, TrustedSigningResources.ManagedIdentityOptionDescription);
    internal Option<string?> TenantIdOption { get; } = new(["-ati", "--azure-tenant-id"], TrustedSigningResources.TenantIdOptionDescription);
    internal Option<string?> ClientIdOption { get; } = new(["-aci", "--azure-client-id"], TrustedSigningResources.ClientIdOptionDescription);
    internal Option<string?> ClientSecretOption { get; } = new(["-acs", "--azure-client-secret"], TrustedSigningResources.ClientSecretOptionDescription);

    internal TokenCredential CreateTokenCredential(InvocationContext context)
    {
       // The logic that is now duplicated in both commands.
    }
}
```

dlemstra: I have update the PR to include these changes. Please let me know what you think about this approach.

dtivel:
> > @dlemstra Isn't the package publicly available in nuget? https://www.nuget.org/packages/Azure.CodeSigning.Sdk
> 
> The nuget.org feed is not included in the `NuGet.Config` file of this project. Wondering if I can add it?

No, see https://github.com/dotnet/arcade/blob/main/Documentation/MirroringPackages.md.  I followed those directions and mirrored the package.

I'm reviewing the rest of your PR.

src/Sign.Cli/TrustedSigningResources.resx
```diff
@@ -0,0 +1,132 @@
...
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="AccountOptionDescription" xml:space="preserve">
+    <value>The Trusted Signing Account name.</value>
+  </data>
+  <data name="CertificateProfileOptionDescription" xml:space="preserve">
+    <value>The Certificate Profile name.</value>
+  </data>
+  <data name="CommandDescription" xml:space="preserve">
+    <value>Use Trusted Signing</value>
```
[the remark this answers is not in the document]
dtivel: Before committing, make sure to run eng\common\CIBuild.cmd to update localizable resource files.

src/Sign.Cli/TrustedSigningResources.resx
```diff
@@ -0,0 +1,132 @@
...
+  </data>
+  <data name="EndpointOptionDescription" xml:space="preserve">
+    <value>The Trusted Signing Account endpoint. The URI value must have a URI that aligns to the region your Trusted Signing Account and Certificate Profile you are specifying were created in during the setup of these resources.</value>
```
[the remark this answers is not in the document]
dlemstra: I borrowed this text from another project. Will update it as you suggested.

src/Sign.SignatureProviders.TrustedSigning/RSATrustedSigning.cs
```diff
@@ -0,0 +1,84 @@
...
+using Azure.CodeSigning.Models;
+
+namespace Sign.SignatureProviders.TrustedSigning
+{
+    internal sealed class RSATrustedSigning : RSA
+    {
+        private readonly CertificateProfileClient _client;
+        private readonly string _accountName;
+        private readonly string _certificateProfileName;
+        private readonly X509Certificate2 _publicKey;
```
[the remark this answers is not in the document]
dlemstra: I named it `_publicKey` instead of `_certificate` to make it clear this only contains the public key and not a private key. I can rename it if you really want me to.

src/Sign.Cli/TrustedSigningResources.resx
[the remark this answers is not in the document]
dlemstra: Made the patch. Those files are automatically updated when I build the project in Visual Studio?

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+                    Response<SignStatus> response = await operation.WaitForCompletionAsync(cancellationToken);
+
+                    byte[] rawData = Convert.FromBase64String(Encoding.UTF8.GetString(response.Value.SigningCertificate));
+                    X509Certificate2Collection collection = [];
+                    collection.Import(rawData);
+
+                    _publicKey = collection[collection.Count - 1];
```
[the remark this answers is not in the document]
dlemstra: This contains the certificate chain of the signing operation that was executed. This always gets the latest certificate in the chain.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        public async Task<X509Certificate2> GetCertificateAsync(CancellationToken cancellationToken)
+        {
+            if (_publicKey is not null)
+            {
+                return new X509Certificate2(_publicKey);
+            }
+
+            await _mutex.WaitAsync();
```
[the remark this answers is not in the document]
dlemstra: That's a good catch. I completely missed that.

src/Sign.SignatureProviders.TrustedSigning/Sign.SignatureProviders.TrustedSigning.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <IsShipping>true</IsShipping>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- This can probably be removed here and in Directory.Packages.props once we upgrade to a different Trusted Signing library -->
```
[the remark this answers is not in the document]
dlemstra: What could I do to make it more clear that we need to check if this can be removed when `Azure.CodeSigning.Sdk` gets replaced with another library?

src/Sign.Cli/AzureAuthOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+            command.AddOption(ClientSecretOption);
+        }
+
+        internal TokenCredential? CreateTokenCredential(InvocationContext context)
+        {
+            bool useManagedIdentity = context.ParseResult.GetValueForOption(ManagedIdentityOption);
+
+            if (useManagedIdentity)
+            {
+                return new DefaultAzureCredential();
```
joelverhagen, line 34: The `--azure-managed-identity` is not exactly an accurate option name. `DefaultAzureCredential` has many behaviors, including token acquisition via the Az CLI. 
https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet

That being said, I think using `DefaultAzureCredential` unlocks a ton of capabilities. For example, GitHub Actions OIDC + the azure/login action light up Az CLI meaning `DefaultAzureCredential` can hook into those ambient credentials, much like a managed identity.

Another scenario which is not clear is user-assigned managed identities. Generally you need to specify the client ID somehow to use a user-assigned managed identity.

My suggestion is to change the option name to `--azure-default-token-credential` and introduce another `--azure-managed-identity` which an option value which is the client ID. This makes the `ManagedIdentityCredential` get used, with the optional `clientId` ctor parameter specified.

joelverhagen: You could also have an `--azure-interactive` switch which enables the interactive mode on `DefaultAzureCredential`
https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential.-ctor?view=azure-dotnet#azure-identity-defaultazurecredential-ctor(system-boolean)

src/Sign.Cli/AzureAuthOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+using Sign.Core;
+
+namespace Sign.Cli
+{
+    internal sealed class AzureAuthOptions
+    {
+        internal Option<bool> ManagedIdentityOption { get; } = new(["-azm", "--azure-managed-identity"], getDefaultValue: () => false, Resources.ManagedIdentityOptionDescription);
+        internal Option<string?> TenantIdOption { get; } = new(["-azt", "--azure-tenant-id"], Resources.TenantIdOptionDescription);
+        internal Option<string?> ClientIdOption { get; } = new(["-azi", "--azure-client-id"], Resources.ClientIdOptionDescription);
+        internal Option<string?> ClientSecretOption { get; } = new(["-azs", "--azure-client-secret"], Resources.ClientSecretOptionDescription);
```
joelverhagen, line 18: Do we really want to top-level this option? There are many better options than client secret these days. IMHO we shouldn't have that option and force the user to, say, use Az CLI with client secret (they do this outside of Sign CLI) then depend on `DefaultAzureCredential` picking up that state.

src/Sign.SignatureProviders.TrustedSigning/Sign.SignatureProviders.TrustedSigning.csproj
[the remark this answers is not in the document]
dlemstra: Figured out why that is happening. The library references this version:
```
.assembly extern Azure.Core
{
  .publickeytoken = (92 74 21 59 E1 2E 44 C8 )                         // .t!Y..D.
  .ver 1:27:0:0
}
```

But the version in the NuGet pacakge is 1.25. So I am pretty sure we can remove this once we switch to another Trusted Signing library.

src/Sign.Cli/AzureAuthOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.CommandLine.Invocation;
+using Azure.Core;
+using Azure.Identity;
+using Sign.Core;
+
+namespace Sign.Cli
+{
+    internal sealed class AzureAuthOptions
+    {
+        internal Option<bool> ManagedIdentityOption { get; } = new(["-azm", "--azure-managed-identity"], getDefaultValue: () => false, Resources.ManagedIdentityOptionDescription);
```
joelverhagen, line 15: I think some of these descriptions mention Key Vault but these don't make sense when the options are used in the `TrustedSigningCommand`

src/Sign.Cli/AzureAuthOptions.cs
dlemstra: See my comment earlier in a response to what @clairernovotny was asking about me moving this out to a separate class.

src/Sign.Cli/TrustedSigningCommand.cs
```diff
@@ -0,0 +1,69 @@
...
+using Sign.Core;
+using Sign.SignatureProviders.TrustedSigning;
+
+namespace Sign.Cli
+{
+    internal sealed class TrustedSigningCommand : Command
+    {
+        internal Option<Uri> EndpointOption { get; } = new(["-tse", "--trusted-signing-endpoint"], TrustedSigningResources.EndpointOptionDescription);
+        internal Option<string> AccountOption { get; } = new(["-tsa", "--trusted-signing-account"], TrustedSigningResources.AccountOptionDescription);
+        internal Option<string> CertificateProfileOption { get; } = new(["-tsc", "--trusted-signing-certificate-profile"], TrustedSigningResources.CertificateProfileOptionDescription);
```
joelverhagen, line 18: Perhaps?

```suggestion
        internal Option<string> CertificateProfileOption { get; } = new(["-tscp", "--trusted-signing-certificate-profile"], TrustedSigningResources.CertificateProfileOptionDescription);
```

src/Sign.Cli/AzureAuthOptions.cs
dlemstra: Thanks for finding that. I will update the documentation once it is clear that this new `AzureAuthOptions` class is a move in the right direction.

src/Sign.Cli/AzureAuthOptions.cs
joelverhagen: I think part of my concern here is that the client secret flow is the only one with explicit support in the options, but other great options exist in the Azure.Identity library. So we're sort of picking a winner which IMO is the least ideal from a security perspective.

src/Sign.Cli/TrustedSigningCommand.cs
dlemstra: Yeah that's a good idea, will update!

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+            ArgumentNullException.ThrowIfNull(tokenCredential, nameof(tokenCredential));
+            ArgumentNullException.ThrowIfNull(endpointUrl, nameof(endpointUrl));
+            ArgumentException.ThrowIfNullOrEmpty(accountName, nameof(accountName));
+            ArgumentException.ThrowIfNullOrEmpty(certificateProfileName, nameof(certificateProfileName));
+
+            _client = new CertificateProfileClient(tokenCredential, endpointUrl);
+            _accountName = accountName;
+            _certificateProfileName = certificateProfileName;
+
+            _logger = serviceProvider.GetRequiredService<ILogger<TrustedSigningService>>();
```
joelverhagen, line 47: From a DI perspective, this is a bit of an antipattern. Ideally, we would inject the logger and not the service provider itself since it's a god object.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+            string accountName,
+            string certificateProfileName)
+        {
+            ArgumentNullException.ThrowIfNull(serviceProvider, nameof(serviceProvider));
+            ArgumentNullException.ThrowIfNull(tokenCredential, nameof(tokenCredential));
+            ArgumentNullException.ThrowIfNull(endpointUrl, nameof(endpointUrl));
+            ArgumentException.ThrowIfNullOrEmpty(accountName, nameof(accountName));
+            ArgumentException.ThrowIfNullOrEmpty(certificateProfileName, nameof(certificateProfileName));
+
+            _client = new CertificateProfileClient(tokenCredential, endpointUrl);
```
joelverhagen, line 43: Should this be a ctor parameter also, perhaps via an interface if available?

dlemstra: There is no interface in this library otherwise I would have used that. I hope that a new library will give us an interface.

joelverhagen: Okay, gotcha. I guess it's a nitpick then. We have a separation of concerns with `TrustedSigningServiceProvider` so we can move the service initialization (`new`) and dependency injection (`GetRequiredService`) out of the ctor here.

I don't feel strongly, but it does seem to better align with my guessed purpose of `TrustedSigningServiceProvider`.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
dlemstra: I kept this in style with the rest of the project but I do think this is a good question. I think that doing `serviceProvider.GetRequiredService<ILogger<TrustedSigningService>>()` inside the `TrustedSigningService` would make more sense. I can change this and do an update to the other SigningProviders in another PR?

src/Sign.Cli/AzureAuthOptions.cs
joelverhagen: I will copy this to the other thread.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
joelverhagen: That sounds good to me. Ideally `IServiceProvider` is passed to as few places as possible. Any code that receives and `IServiceProvider` can do ANYTHING. So it makes the code harder to understand and predict.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
dlemstra: As mentioned in my other comment I do think it's a good idea to remove the `GetRequiredService` call here. We could introduce a wrapper around the `CertificateProfileClient` class but I also wonder if we need to make that investment now when we will replace this with another TrustedSigning library once this becomes available.

joelverhagen: You could resolve `CertificateProfileClient` in the call site of this ctor and simply pass in the implementation as a parameter. I only call this out because I've generally been served well by following DI patterns as closely as possible even in CLI contexts. But I agree we can leave this as-is and re-evaluate as the dependent library changes. I'll leave it up to you on what to do.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
dlemstra: I have update the PR to make this change. And I will create another PR to make the changes to the other SigningProviders once this has been merged.

src/Sign.Cli/TrustedSigningCommand.cs
joelverhagen: Ah, sorry I missed this thread. I'll summarize my [other thoughts](https://github.com/dotnet/sign/pull/716#discussion_r1644898954) here and add additional things.

Trusted Signing and Key Vault both are Azure-only things so `--azure` prefix on the option names seemed redundant. But I don't feel strongly.

I like the extensibility of `--azure-auth-type` but we would need some additional conventions when parameters are needed or are optional for a given auth type. For example, would `--azure-client-id` be used only for the `client-secret` type or also by `managed-identity` for the user-assigned managed identity case? What if `client-id` in one case needs a different description than another case? We could do a verbose `--azure-client-secret-client-id` and `--azure-managed-identity-client-id` perhaps. Given auth types are mutually exclusive, I think it's better to re-use the options and keep the descriptions sufficiently descriptive for all use cases.

I am not convinced we should add the `client-secret` type. I think we could describe how to work around the Sign CLI limitation with Az CLI + `DefaultAzureCredential` but gently urge folks to just not use that and at least try to find a better option. Said another way, I don't think we should make it easy to user client secret but still allow it via workaround that the user needs to put a bit more work into.

**In summary, my recommendation** (feel free to pick and choose or say no 😁)

IMO I think `--interactive` should default to `false` so that people don't accidentally prompt UI in a non-interactive place (e.g. CI). In other words, I think CLI tools that are reasonable to use in CIs should assume non-interactive unless specified otherwise. I believe `dotnet restore` has this default, you need to opt in with `--interactive` ([docs](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-restore#options)).

Having explicit managed identity support seems not important for the first pass since it can be done with `DefaultAzureCredential`. So I think we can skip `--azure-auth-type` for now and have it essentially always be `default`. If you want `azure-auth-type` right now, then `default` (mapping to `DefaultAzureCredential`) and `managed-identity`  (mapping to `ManagedIdentityCredential`) seem like the right first two to add.

Other recommendations:

`--interactive` should be `--azure-interactive-auth` to be more specific on what exactly is interactive

`--azure-client-id` should be used for both clientId in `ManagedIdentityCredential` and `DefaultAzureCredentialOptions.ManagedIdentityClientId` (used for both `default` and `managed-identity` types)

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
dlemstra: I hope that I can inject a client or a factory to create the client (because we need to specify `tokenCredential` and `endpointUrl`). That is why I have hidden this in the constructor for now. I will leave it up to the team to make the final decision.

src/Sign.Cli/AzureAuthOptions.cs
joelverhagen: This is a duplicate of https://github.com/dotnet/sign/pull/716#discussion_r1644941441.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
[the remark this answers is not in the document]
dtivel: Perhaps add a comment:

```C#
// This should contain the certificate chain in root->leaf order.
```

`SigningCertificate` on line 76 suggests there's one certificate not many, let alone an ordered set of certificates.

src/Sign.Cli/TrustedSigningResources.resx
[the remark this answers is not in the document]
dtivel: I don't think so, but I may be wrong.  (Try it.)  As of writing this, the localizable resource files don't have your change.  See https://github.com/dotnet/sign/pull/716/files/39b9d205abd6d374c22a261b2da13ce566ef45d9#diff-e2bb297d4ce9b440839fe3855706849a2114a59f3a4dabe333eaee2d6e139929R16

I think you need to run eng\common\CIBuild.cmd before committing.

src/Sign.SignatureProviders.TrustedSigning/RSATrustedSigning.cs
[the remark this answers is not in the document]
dlemstra: Okay will do.

src/Sign.SignatureProviders.TrustedSigning/Sign.SignatureProviders.TrustedSigning.csproj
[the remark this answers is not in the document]
dtivel: I think I understand.  Azure.CodeSigning.Sdk 0.1.96 pulls in Azure.Core 1.27.0.  We want Azure.Core 1.38.0 or later.

It wasn't clear to me if by "once we switch to another Trusted Signing library" if there's a replacement for Azure.CodeSigning.Sdk we should be looking forward to or not.  It sounds like we just need a newer version of Azure.CodeSigning.Sdk that references Azure.Core 1.38.0 or later.  Is that a correct summary?

If yes, then how about:

```suggestion
    <!-- When Azure.CodeSigning.Sdk references Azure.Core 1.38.0 or later we should be able to remove the explicit reference to Azure.Core here and in Directory.Packages.props. -->
```

Months from now, if I read that I'll understand the intent.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
clairernovotny: seems like we shouldn't be disposing that cert then, at least not at that scope?

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+
+            await _mutex.WaitAsync(cancellationToken);
+            try
+            {
+                if (_certificate is null)
+                {
+                    Stopwatch stopwatch = Stopwatch.StartNew();
+
+                    _logger.LogTrace(Resources.FetchingCertificate);
```
dtivel, line 70: `Resources.FetchingService` returns:

`Fetching certificate from Azure Key Vault.`

We could either flavor this message for each certificate provider or make the message more generic (e.g.:  `Fetching certificate.`).

The real problem is that Sign.SignatureProviders.TrustedSigning uses Sign.Core.Resources.  While this is temporarily OK, Sign.SignatureProviders.TrustedSigning should not directly reference Sign.Core.Resources; each should have its own resource strings.

Because we lack an extensibility model it's passable to reference Sign.Core.Resources for now.  Making the message more generic is OK now, but the better solution would be to create a separate string resources.

I notice that [`CertificateStoreService`](https://github.com/dotnet/sign/blob/db11f797c7c1ba2d133a63474a0d458f9dc178df/src/Sign.SignatureProviders.CertificateStore/CertificateStoreService.cs#L115) has the same problem.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
dtivel: This feedback and others like it is out of scope for this PR.  @clairernovotny, your concerns are based on an existing design, which @dlemstra is following correctly.

It is a best practice to dispose of `IDisposable` instances for which you've taken ownership of and you no longer need.  @dlemstra's changes follow existing patterns.  The scope of this PR should not include the suggested design change.

In the future, we could redesign signers to reuse `X509Certificate2` instances as much as possible.  For example, an `X509Certificate2` instance could be passed to `IDataFormatSigner.SignAsync(...)`.  The instance would be owned and disposed of at a more outer scope.  There may be alternatives.

Also, while performance is important, so are many other things, like maintainability, readability, etc.  Measuring the performance benefit of such a change would be very valuable in cost/benefit decisions.  Reusing `X509Certificate2` instances probably improves overall performance (by virtue of fewer allocations and GC's), but without measurements, it's hard to say if it'll be a significant benefit.

For example, for signing a NuGet package and the assemblies inside of it, it looks like --- from a cursory glance --- we'll instantiate and dispose of 2 `X509Certificate2` objects in the service of `GetCertificateAsync(...)` calls.  A design change that would reuse `X509Certificate2` instances would bring that down from 2 to 1.  On the surface (read:  without actual measurements), all of that redesign work to eliminate 1 instantiation seems like not a good use of perf optimization time.  IMO, a better use of time would be to profile Sign CLI end-to-end and look for low-hanging fruit.

@clairernovotny, please open a new issue if you feel strongly about this.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
dlemstra: That's why I thought it was safe to copy this resource. I haven't looked at the content of those message as you and Joel noticed. I will add a resource file for the Trusted Signing. For the other two SignatureProviders I will make those changes in the other PR that changes the DI of the service.

src/Sign.SignatureProviders.TrustedSigning/Sign.SignatureProviders.TrustedSigning.csproj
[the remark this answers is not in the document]
dlemstra: I have updated the comment to make it more clear why we need this reference.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
dlemstra: I think we should do that at a later moment when a library that contains a call for this becomes available.

src/Sign.Cli/TrustedSigningCommand.cs
dtivel: https://github.com/dotnet/sign/issues/561 is an example of how `DefaultAzureCredential` used an authentication strategy other than managed identity.  Having a managed identity-specific option does not make sense to me.  It should be simply "use Azure authentication."  There are many authentication strategies for Azure, and we should not implement support for them one by one.

vsce, which is the extension publishing tool for VS Code, has an [`--azure-credential`](https://github.com/microsoft/vscode-vsce/blob/33daacd8595c663db531252bb2ba624111426012/src/main.ts#L188) option that uses [`DefaultAzureCredential`](https://github.com/microsoft/vscode-vsce/blob/33daacd8595c663db531252bb2ba624111426012/src/util.ts#L56-L57).

It makes more sense to me to replace `--azure-managed-identity` with `--azure-credential` (or `--azure-authentication` or whatever) and remove the other Azure options.

The behavior of `DefaultAzureCredential` can be customized with `ExcludeXXXCredential` options to opt out of specific unwanted strategies.

dlemstra: I even wonder if we still need the `--azure-credential` option. What if leaving it out would always create a `DefaultAzureCredential` but we keep `--azure-tenant-id`, `--azure-client-id`, `--azure-client-secret`. When those options are specified we return a `ClientSecretCredential` instead so people who use a client-id/client-secret will only need to rename their arguments. We could even keep the old `--azure-key-vault-*` names and then report a warning they have been renamed and should no longer be used. And we could also add options to specify `--exclude-environment-credential` and the rest of those exclude options. We could then make `--exclude-interactive-browser-credential` default to `false` to make sure users don't get a "browser interaction" in their CI. I think this approach would make things a lot easier?

joelverhagen: Most of the options discussed here appear to be a matter of taste (IMO). I think the only "controversial" decisions are:

- Do we support client secret at all? My opinion: **no**. They can use Az CLI client secret auth (`az login --password`) + `DefaultAzureCredential` to workaround.
- Do we default interactive or not? My opinion (and appearing consensus here): **no**
- Do we have breaking changes with the existing `--azure-key-vault-*` names? My opinion (pretty worthless, I don't have the context): **make the breaking change, we are prerelease on the Sign CLI**. Warning seems like a nice thing too but is extra work.

Other options like whether to have explicit Managed Identity support or whether to opt-in or opt-out modes of `DefaultAzureCredential` seem like we can make a good guess (choose something that is consistent and looks like enough) and listen to customers.

> The behavior of `DefaultAzureCredential` can be customized with `ExcludeXXXCredential` options to opt out of specific unwanted strategies.

I am not a huge can of this personally since, if you know exactly what credential you want, you need to keep playing whack-a-mole with the `Exclude*` flags as Azure.Identity adds more modes to `DefaultAzureCredential`. I think the idea of @dlemstra to have something like `--azure-auth-type` allows us to add `environment-credential` or `visual-studio` types.

Said another way, I think there are two main categories of users (if I were to guess): "embrace the magic" (use `DefaultAzureCredential` with default options) or "I know what I'm doing" (use a specific `TokenCredential` implementation not `DefaultAzureCredential`). The `--exclude` or opt-out pattern is somewhere in between those two camps of users which feels awkward to me.

Whether or not to have `--azure-managed-identity` (or the equivalent `--azure-auth-type managed-identity` idea) is guesswork for me. I think it's not worth it since it wouldn't be useful in GitHub Actions where I hear the most users asking for it (I am not plugged in like Damon and Claire though to feedback). Do we really expect people to be signing from Azure compute directly? This seems more like a CI or devbox tool where `DefaultAzureCredential` (or an explicit `--azure-auth-type interactive-browser`) would be more useful.

dlemstra: I can agree with Joel their answers to the questions that we need to answer. We would then only need a new `--azure-auth-type` option and allow an explicit opt-in to one of the types (e.g. `EnvironmentCredential`, `ManagedIdentityCredential`) or use the default value `DefaultAzureCredential`. We should probably also allow specifying the rest of the `DefaultAzureCredentialOptions`. I am not sure if we would want to do that now or at this at a later moment? Maybe adding just `--azure-tenant-id` to set `DefaultAzureCredentialOptions.TenantId` would be sufficient for now?

This would be a breaking change for all the users but as Joel said we should be allowed to do that in a prerelease? I do think there should be some really good release notes that explain these changes and also explain why these changes were made?

While writing this reply I am also wondering if `--azure-credential` would be a better name than `--azure-auth-type`?

dtivel: I think we should take up further changes to Azure authentications in a separate issue/PR with a succinct proposal of the changes.

I love the idea of removing all existing options and Azure Key Vault and Trusted Signing signature providers using `DefaultAzureCredential` implicitly by default.  Also, having an option that enables a user to opt in to a specific authentication strategy instead of the trial-and-error sequence of authentication strategies that `DefaultAzureCredential` uses.

dtivel (approved):

src/Sign.Cli/TrustedSigningCommand.cs
dlemstra: Sounds like a good idea to discuss this in a separate PR. I have a proof of concept ready so I can create a PR for that once this is merged. I do think we should make `DefaultAzureCredential` the default instead and that would mean we can make the `-kvm` option obsolete. We will use `ClientSecretCredential` when `TenantIdOption`, `ClientIdOption` and `ClientSecretOption` are specified otherwise we default to `DefaultAzureCredential`. I have updated the PR with this proposal.

src/Sign.Cli/AzureAuthOptions.cs
dlemstra: I have updated the documentation.

clairernovotny (approved):

Issue #724, opened 2024-06-20
CLI:  simplify Azure authentication


dtivel:
[This thread](https://github.com/dotnet/sign/pull/716#discussion_r1640156057) generated some ideas on how we can improve the CLI user experience for Azure authentication, which is currently used by Azure Key Vault signing.  In the near future, it will also be used by Trusted Signing signing.

## Background

Sign CLI's Azure Key Vault signing exposes [4 authentication-specific options](https://github.com/dotnet/sign/blob/db11f797c7c1ba2d133a63474a0d458f9dc178df/src/Sign.Cli/AzureKeyVaultCommand.cs#L20-L23).  The original intent was to support 2 authentication strategies:

- tenant ID + client ID + client secret
- managed identity

## Problem statement

Managed identity support is provided through Azure SDK for .NET's [`DefaultAzureCredential`](https://learn.microsoft.com/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet) class.  This class actually supports 10 authentication strategies, which are tried in a fixed order until one succeeds.  In most cases, this may work like magic.  However, there are several problems with this solution.

1.  `--azure-key-vault-managed-identity` doesn't actually guarantee we'll use the managed identity authentication strategy.  Authentication can fail because `DefaultAzureCredential` succeeded in obtaining a token with an earlier (unwanted) authentication strategy.  [This really happens.](https://github.com/microsoft/vscode-vsce/issues/976)
1.  There's no clear support for other authentication strategies (like Azure CLI).  You could use the `--azure-key-vault-managed-identity` option, set a bunch of [`ExcludeXXXCredential`](https://learn.microsoft.com/dotnet/api/azure.identity.defaultazurecredentialoptions?view=azure-dotnet#properties) options for all the options before Azure CLI authentication strategy, but that's a hacky, terrible user experience.
1.  From a CLI UX perspective, to formally support other authentication strategies (like Azure CLI), we'd need to expose an option for it and validate the various combinations of options.

## Proposed solution

Remove the 4 authentication-specific options.  By default, Azure Key Vault signing will require no Azure authentication options.  It will support all authentication strategies provided by `DefaultAzureCredential` in the order that class provides.

[...]

PR #725, opened 2024-06-24
Added --azure-credential-type that can be used to specify which Azure credential type should be enabled.

dlemstra:
This pull requests adds the `--azure-credential-type` option that was discussed in #724. The easiest to start with was `environment` but it is build so we can add

### DIFF
--- a/src/Sign.Cli/AzureCredentialOptions.cs
+++ b/src/Sign.Cli/AzureCredentialOptions.cs
@@ -15,7 +15,7 @@ namespace Sign.Cli
     {
         internal Option<string?> CredentialTypeOption { get; } = new Option<string?>(["-act", "--azure-credential-type"], Resources.CredentialTypeOptionDescription).FromAmong(
             AzureCredentialType.Environment);
-        internal Option<bool?> ManagedIdentityOption { get; } = new(["-kvm", "--azure-key-vault-managed-identity"], Resources.ManagedIdentityOptionDescription);
+        internal Option<bool?> ManagedIdentityOption { get; } = new(["-kvm", "--azure-key-vault-managed-identity"], Resources.ManagedIdentityOptionDescription) { IsHidden = true };
         internal Option<string?> TenantIdOption { get; } = new(["-kvt", "--azure-key-vault-tenant-id"], Resources.TenantIdOptionDescription);
         internal Option<string?> ClientIdOption { get; } = new(["-kvi", "--azure-key-vault-client-id"], Resources.ClientIdOptionDescription);
         internal Option<string?> ClientSecretOption { get; } = new(["-kvs", "--azure-key-vault-client-secret"], Resources.ClientSecretOptionDescription);
--- a/src/Sign.Cli/AzureCredentialOptions.cs
+++ b/src/Sign.Cli/AzureCredentialOptions.cs
@@ -5,6 +5,7 @@
 using System.CommandLine;
 using System.CommandLine.Invocation;
 using System.CommandLine.IO;
+using System.CommandLine.Parsing;
 using Azure.Core;
 using Azure.Identity;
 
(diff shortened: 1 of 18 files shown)


src/Sign.Cli/AzureCredentialOptions.cs
```diff
@@ -5,26 +5,49 @@
...
         internal void AddOptionsToCommand(Command command)
         {
+            command.AddOption(CredentialTypeOption);
             command.AddOption(ManagedIdentityOption);
             command.AddOption(TenantIdOption);
             command.AddOption(ClientIdOption);
             command.AddOption(ClientSecretOption);
         }
 
+        internal DefaultAzureCredentialOptions CreateDefaultAzureCredentialOptions(ParseResult parseResult)
```
dtivel, line 32: This will get complicated as we add explicit support for specific authentication strategies.

I was envisioning something simpler and more scalable:

```C#
if (string.Equals(AzureCredentialType.Environment, credentialType, StringComparison.OrdinalIgnoreCase))
{
    return new EnvironmentCredential();
}
else
{
    return new DefaultAzureCredential();
}
```

dtivel: After looking at documentation, I see that [`DefaultAzureCredential`'s default constructor is only intended for mocking](https://learn.microsoft.com/dotnet/api/azure.identity.defaultazurecredential.-ctor?view=azure-dotnet#overloads).  We shouldn't be using it here.

`EnvironmentCredential` does have a default constructor that is fine to use here.

However, [`ManagedIdentityCredential`'s default constructor is also intended for mocking](https://learn.microsoft.com/dotnet/api/azure.identity.managedidentitycredential?view=azure-dotnet#constructors).

I'm not sure yet what it would look like for us to construct a `ManagedIdentityCredential` with non-default options.  (I haven't looked at what `DefaultAzureCredential` does.)

dlemstra: With `new DefaultAzureCredential()` we would be using this constructor: `public DefaultAzureCredential(bool includeInteractiveCredentials = false)`. I thought that the advantage of `DefaultAzureCredential` with the options would be that we would not need to construct each of the `Credential` classes and just use the logic that is inside `DefaultAzureCredential`. Otherwise we would probably create something that looks a lot like the implementation behind `DefaultAzureCredential`. And we would need to maintain that then. With this strategy we allow a user to specify which one the credential types they would want to use. And we can still change this in the future if we really need to?

src/Sign.Cli/AzureCredentialOptions.cs
```diff
@@ -5,26 +5,49 @@
...
+        {
+            DefaultAzureCredentialOptions options = new();
+
+            string? credentialType = parseResult.GetValueForOption(CredentialTypeOption);
+            if (credentialType is not null)
+            {
+                options.ExcludeAzureCliCredential = true;
+                options.ExcludeAzureDeveloperCliCredential = true;
+                options.ExcludeAzurePowerShellCredential = true;
+                options.ExcludeEnvironmentCredential = credentialType != AzureCredentialType.Environment;
```
dtivel, line 42: We're case insensitive with other option values (like file and timestamp digest algorithms), so I think we should be case sensitive here too.

src/Sign.Cli/AzureCredentialType.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+namespace Sign.Cli
+{
+    internal static class AzureCredentialType
+    {
+        public const string Environment = "environment";
```
dtivel, line 9: Should we add `managed-identity` now?  @clairernovotny, thoughts?

dlemstra: Might be better to do this in separate PR's?

src/Sign.Cli/AzureCredentialOptions.cs
dlemstra: We are now not case sensitive because `FromAmong` is used. They need to be entered in the same casing as specified:

```
Argument 'eNviroNment' not recognized. Must be one of:
        'environment'
```

Would you want me to change this or are you okay with me forcing the user to do this in lowercase?

src/Sign.Cli/Resources.resx
```diff
@@ -135,6 +135,9 @@
   <data name="CodeCommandDescription" xml:space="preserve">
     <value>Sign binaries and containers.</value>
   </data>
+  <data name="CredentialTypeOptionDescription" xml:space="preserve">
+    <value>Azure credential type that will be used. This defaults to all types except interactive-browser, shared-token and visual-studio-code.</value>
```
dtivel, line 139: I don't think we need to hyphenate, since these are explicitly not supported values.

```suggestion
    <value>Azure credential type that will be used. This defaults to all types except interactive browser, shared token, and Visual Studio Code.</value>
    <
```

dlemstra: Thanks for improving this message!

src/Sign.Cli/AzureCredentialOptions.cs
dtivel: OK, let's leave it as-is until we get more feedback.  The default error message is clear.

dtivel (approved):

PR #726, opened 2024-06-25
Added --azure-tenant-id option and made all the --azure-key-vault options obsolete.

dlemstra:
This PR makes all the `--azure-key-vault` obsolete and hidden. When the options are hidden they can still be used but they will no longer show up in the help. And this PR also introduces an option to set the `TenantId` property of the `DefaultAzureCredential`.

### DIFF
--- a/src/Sign.Cli/AzureCredentialOptions.cs
+++ b/src/Sign.Cli/AzureCredentialOptions.cs
@@ -16,23 +16,31 @@ namespace Sign.Cli
         internal Option<string?> CredentialTypeOption { get; } = new Option<string?>(["--azure-credential-type", "-act"], Resources.CredentialTypeOptionDescription).FromAmong(
             AzureCredentialType.AzureCli,
             AzureCredentialType.Environment);
-        internal Option<bool?> ManagedIdentityOption { get; } = new(["--azure-key-vault-managed-identity", "-kvm"], Resources.ManagedIdentityOptionDescription) { IsHidden = true };
-        internal Option<string?> TenantIdOption { get; } = new(["--azure-key-vault-tenant-id", "-kvt"], Resources.TenantIdOptionDescription);
-        internal Option<string?> ClientIdOption { get; } = new(["--azure-key-vault-client-id", "-kvi"], Resources.ClientIdOptionDescription);
-        internal Option<string?> ClientSecretOption { get; } = new(["--azure-key-vault-client-secret", "-kvs"], Resources.ClientSecretOptionDescription);
+        internal Option<string?> TenantIdOption { get; } = new(["--azure-tenant-id", "-ati"], Resources.TenantIdOptionDescription);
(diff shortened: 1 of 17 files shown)


clairernovotny (approved):

src/Sign.Cli/AzureCredentialOptions.cs
```diff
@@ -13,26 +13,34 @@ namespace Sign.Cli
...
     {
-        internal Option<string?> CredentialTypeOption { get; } = new Option<string?>(["-act", "--azure-credential-type"], Resources.CredentialTypeOptionDescription).FromAmong(
+        internal Option<string?> CredentialTypeOption { get; } = new Option<string?>(["--azure-credential-type", "-act"], Resources.CredentialTypeOptionDescription).FromAmong(
             AzureCredentialType.Environment);
-        internal Option<bool?> ManagedIdentityOption { get; } = new(["-kvm", "--azure-key-vault-managed-identity"], Resources.ManagedIdentityOptionDescription) { IsHidden = true };
-        internal Option<string?> TenantIdOption { get; } = new(["-kvt", "--azure-key-vault-tenant-id"], Resources.TenantIdOptionDescription);
-        internal Option<string?> ClientIdOption { get; } = new(["-kvi", "--azure-key-vault-client-id"], Resources.ClientIdOptionDescription);
-        internal Option<string?> ClientSecretOption { get; } = new(["-kvs", "--azure-key-vault-client-secret"], Resources.ClientSecretOptionDescription);
+        internal Option<string?> TenantIdOption { get; } = new(["--azure-tenant-id", "-ati"], Resources.TenantIdOptionDescription);
```
dtivel, line 18: Why is this necessary?

For example, [`EnvironmentCredential` picks up the tenant ID from an environment variable](https://learn.microsoft.com/dotnet/api/azure.identity.environmentcredential?view=azure-dotnet#definition).  Is there another credential type that needs the tenant ID passed in some other way?

test/Sign.Cli.Test/AzureCredentialOptionsTests.cs
```diff
@@ -37,57 +37,87 @@ public void CredentialTypeOption_Always_IsNotRequired()
         }
 
         [Fact]
-        public void ManagedIdentityOption_Always_HasArityOfZeroOrOne()
+        public void TenantIdOption_Always_HasArityOfZeroOrOne()
```
dtivel, line 40:
```suggestion
        public void TenantIdOption_Always_HasArityOfExactlyOne()
```

dlemstra: Thanks for noticing those copy paste mistakes.

src/Sign.Cli/AzureCredentialOptions.cs
dlemstra: It is possible that you login with `az login` without specifying a tenant. And according to the comment of the property this can also be useful when you login to one tenant but want to use another tenant:

```
The ID of the tenant to which the credential will authenticate by default. If not specified, the credential will authenticate to any requested tenant, and will default to the tenant to which the chosen authentication method was originally authenticated.
```

I thought it would be useful to have this option.

dtivel (approved):

PR #730, opened 2024-06-25
Added managed-identity credential type.

dlemstra:
This PR adds the managed-identity credential type. This also adds two new option that are specific for a managed identity in the `DefaultAzureCredentialOptions`. I can take those out if the team thinks that it is too early to add those.


clairernovotny:
Seems reasonable to expose the full range of managed identity options that the library provides so that we're not blocking functionality

clairernovotny (approved):

Issue #735, opened 2024-07-02
Error signing clickonce with Trusted Signing


JaapMosselman:
``I am trying to use the new Trusted Signing option of dotnet sign which @dlemstra added.
I am able to sign a msi, works fine, but have problems with clickonce.

I use this powershell in my devops pipeline:
              .\sign code trusted-signing `
                "$applicationFile" `
                --base-directory "$applicationDir" `
                --application-name "...." `
                --publisher-name "...." `
                --trusted-signing-endpoint "...." `
                --trusted-signing-account "...." `
                --trusted-signing-certificate-profile "...." `
                --managed-identity-client-id "...."

I get this error:
`fail: Sign.Core.ISigner[0]
      Specified method is not supported.
      System.ApplicationException: Specified method is not supported.
       ---> System.NotSupportedException: Specified method is not supported.
         at System.Deployment.Internal.CodeSigning.SignedCmiManifest2.AuthenticodeSignLicenseDom(XmlDocument licenseDom, CmiManifestSigner2 signer, String timeStampUrl, Boolean disallowMansignTimestampFallback) in /_/src/Sign.Core/Native/mansign2.cs:line 676
         at System.Deployment.Internal.CodeSigning.SignedCmiManifest2.Sign(CmiManifestSigner2 signer, String timeStampUrl, Boolean disallowMansignTimestampFallback) in /_/src/Sign.Core/Native/mansign2.cs:line 357
         at Sign.Core.ManifestSigner.Sign(FileInfo file, X509Certificate2 certificate, RSA rsaPrivateKey, SignOptions options) in /_/src/Sign.Core/DataFormatSigners/ManifestSigner.cs:line 42
         --- End of inner exception stack trace ---
         at Sign.Core.ManifestSigner.Sign(FileInfo file, X509Certificate2 certificate, RSA rsaPrivateKey, SignOptions options) in /_/src/Sign.Core/DataFormatSigners/ManifestSigner.cs:line 52
         at Sign.Core.ClickOnceSigner.SignCoreAsync(String args, FileInfo file, RSA rsaPrivateKey, X509Certificate2 certificate, SignOptions options) in /_/src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs:line 220
[...]

PR #736, opened 2024-07-02
Fix ClickOnce signing using Trusted Signing

dlemstra:
This PR fixes the issue that was reported in #735. There was an explicit cast to `RSAKeyVault` but this can be changed to `RSA` instead. I have updated a unit test to hit this "new" path and also removed an unnecessary duplicate check.

### DIFF
--- a/src/Sign.Core/Native/mansign2.cs
+++ b/src/Sign.Core/Native/mansign2.cs
@@ -5,7 +5,6 @@
 #define RUNTIME_TYPE_NETCORE
 
 using System.Diagnostics.CodeAnalysis;
-using System.IO;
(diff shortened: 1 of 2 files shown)


dtivel (approved):

dtivel:
@clairernovotny, do you want to review this?

JaapMosselman:
Any estimation for the timeframe this PR will be released?
Whould be nice if I don't have to use my own version.

dtivel:
@JaapMosselman, https://www.nuget.org/packages/sign/0.9.1-beta.24361.2 has the fix.  Happy signing!

PR #740, opened 2024-07-11
refactor get certificate async function to use get cert chain operation

Jaxelr:
Hey,

Given that I added the get certificate chain operation to the nuget.org [azure.codesigning.sdk](https://www.nuget.org/packages/Azure.CodeSigning.Sdk/0.1.106). I thought it would be appropriate, given that trusted signing charges per signature, to create this PR.  This also improves the performance of the signing step.

image

image

Aside from the usual live test, I don't see any integration tests so I didn't dared adding those, but if you would like me to, please let me know.

cc: @dlemstra

### DIFF
--- a/src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
+++ b/src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
@@ -5,8 +5,6 @@
 using System.Diagnostics;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
-using System.Text;
-using Azure;
 using Azure.CodeSigning;
 using Azure.CodeSigning.Models;
 using Azure.Core;
@@ -69,12 +67,13 @@ namespace Sign.SignatureProviders.TrustedSigning
 
                     _logger.LogTrace(Resources.FetchingCertificate);
 
-                    CertificateProfileSignOperation operation = await _client.StartSignAsync(_accountName, _certificateProfileName, _emptyRequest, cancellationToken: cancellationToken);
-                    Response<SignStatus> response = await operation.WaitForCompletionAsync(cancellationToken);
+                    var response = await _client.GetSignCertificateChainAsync(_accountName, _certificateProfileName, cancellationToken: cancellationToken);
 
-                    byte[] rawData = Convert.FromBase64String(Encoding.UTF8.GetString(response.Value.SigningCertificate));
+                    var rawData = new MemoryStream();
+                    response.Value.CopyTo(rawData);
+
                     X509Certificate2Collection collection = [];
-                    collection.Import(rawData);
+                    collection.Import(rawData.ToArray());
 
                     // This should contain the certificate chain in root->leaf order.
                     _certificate = collection[collection.Count - 1];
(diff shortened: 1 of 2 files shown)


src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -69,12 +67,13 @@ public async Task<X509Certificate2> GetCertificateAsync(CancellationToken cancel
 
                     _logger.LogTrace(Resources.FetchingCertificate);
 
-                    CertificateProfileSignOperation operation = await _client.StartSignAsync(_accountName, _certificateProfileName, _emptyRequest, cancellationToken: cancellationToken);
-                    Response<SignStatus> response = await operation.WaitForCompletionAsync(cancellationToken);
+                    var response = await _client.GetSignCertificateChainAsync(_accountName, _certificateProfileName, cancellationToken: cancellationToken);
```
dlemstra, line 70: This project doesn't use `var` when the type is not directly visible on the right side.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -69,12 +67,13 @@ public async Task<X509Certificate2> GetCertificateAsync(CancellationToken cancel
...
+
+                    var rawData = new MemoryStream();
```
dlemstra, line 72:
```suggestion
                    MemoryStream rawData = new();
```

Jaxelr: I'm going to refactor this to use byte[] directly from the stream.

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
Jaxelr: Shouldn't that be covered by the editorconfig? I saw no feedback. I updated the PR to use the explicit type. Thanks!

dlemstra (approved):
Thanks for adding this!

src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs
```diff
@@ -69,10 +68,11 @@ public async Task<X509Certificate2> GetCertificateAsync(CancellationToken cancel
 
                     _logger.LogTrace(Resources.FetchingCertificate);
 
-                    CertificateProfileSignOperation operation = await _client.StartSignAsync(_accountName, _certificateProfileName, _emptyRequest, cancellationToken: cancellationToken);
-                    Response<SignStatus> response = await operation.WaitForCompletionAsync(cancellationToken);
+                    Response<Stream> response = await _client.GetSignCertificateChainAsync(_accountName, _certificateProfileName, cancellationToken: cancellationToken);
```
dtivel, line 71: Dispose of the stream?

Jaxelr: Added Dispose(), thanks!

dtivel (approved):

PR #755, opened 2024-08-09
Correct creation of the CryptographyClient in the KeyVaultService.

dlemstra:
This corrects the implementation of the creation of the CryptographyClient. The `KeyClient` that was created used the `KeyId` instead of the url of the keyvault here: `KeyClient keyClient = new(certificateWithPolicy.KeyId, _tokenCredential);`. And that was then used to create a `CryptographyClient` with an incorrect url.

I have fixed this by creating the `CryptographyClient` without using the `KeyClient` because we already get this information when the public key is retrieved. I have also removed the `_task` field to make it possible to pass the `cancellationToken` to the `GetCertificateAsync` call.


dtivel (approved):

dtivel:
Looks good, @dlemstra.  Thanks for the PR.

PR #758, opened 2024-08-28
Better manifest file detection

jackmtpt:
This PR changes the ClickOnce signer to better detect which .dll.manifest or .exe.manifest Application manifest file to sign. It reads the supplied .vsto/.application Deployment manifest file and parses out the name of the Application manifest that's referenced, and then passes that to `mage` for updating.

### DIFF
--- a/src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
+++ b/src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
@@ -5,6 +5,7 @@
 using System.Globalization;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
+using System.Xml;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileSystemGlobbing.Abstractions;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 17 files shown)


src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
```diff
@@ -273,5 +274,53 @@ public void CopySigningDependencies(FileInfo deploymentManifestFile, DirectoryIn
...
+            }
+
+            // the codebase attribute can be a relative file path (e.g. Application Files\MyApp_1_0_0_0\MyApp.dll.manifest) or
+            // a URI (e.g. https://my.cdn.com/clickonce/MyApp/ApplicationFiles/MyApp_1_0_0_0/MyApp.dll.manifest) so we need to
+            // handle both cases and extract just the file name part.
+            //
+            // we only try and parse absolute uris, because a relative uri can just be treated like a file path for our purposes
+            if (Uri.TryCreate(applicationManifestFileNameAttribute.Value, UriKind.Absolute, out var uri))
+            {
+                Path.GetFileName(uri.LocalPath); // works for http(s) and file:// uris
```
dtivel, line 320: There is a bug here.  I think you meant:

```suggestion
                return Path.GetFileName(uri.LocalPath); // works for http(s) and file:// uris
```

Seems worth adding a unit test for this case.  I think you can convert your new, existing unit test into a theory and parameterize `codebase` with both scenarios.

src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
```diff
@@ -126,8 +127,8 @@ await Parallel.ForEachAsync(files, _parallelOptions, async (file, state) =>
 
                     // Inner files are now signed
                     // now look for the manifest file and sign that if we have one
-
-                    FileInfo? manifestFile = filteredFiles.SingleOrDefault(f => ".manifest".Equals(f.Extension, StringComparison.OrdinalIgnoreCase));
+                    var appManifestFromDeploymentManifest = GetApplicationManifestForDeploymentManifest(file);
+                    FileInfo? manifestFile = filteredFiles.SingleOrDefault(f => f.Name.Equals(appManifestFromDeploymentManifest, StringComparison.OrdinalIgnoreCase));
```
dtivel, line 131: This is avoidable if `appManifestFromDeploymentManifest` is null.

Also, can we rename `manifestFile` to `applicationManifestFile` since now we're dealing with potentially many different kinds of manifest files.  We had deployment and application manifests, but now we're dealing with other manifests (like assembly manifests).

src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
```diff
@@ -273,5 +274,53 @@ public void CopySigningDependencies(FileInfo deploymentManifestFile, DirectoryIn
                 }
             }
         }
+
+        /// <summary>
+        /// Try and find the application manifest (.manifest) file from a clickonce application manifest (.application / .vsto
```
dtivel, line 279: Shouldn't it be:

```suggestion
        /// Try and find the application manifest (.manifest) file from a clickonce deployment manifest (.application / .vsto
```

src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
```diff
@@ -273,5 +274,53 @@ public void CopySigningDependencies(FileInfo deploymentManifestFile, DirectoryIn
...
+        /// </summary>
+        /// <param name="deploymentManifest"></param>
+        /// <returns>A string containing the file name of the Application manifest, or null if it couldn't be found.</returns>
+        /// <exception cref="InvalidDataException"></exception>
+        private string? GetApplicationManifestForDeploymentManifest(FileInfo deploymentManifest)
+        {
+            var xml = new XmlDocument();
+            xml.Load(deploymentManifest.FullName);
+            // there should only be a single result here, if the file is a valid clickonce manifest.
+            var dependentAssemblies = xml.GetElementsByTagName("dependentAssembly");
```
dtivel, line 292: We prefer using the explicit type if the implied type is not super obvious.  It will facilitate understanding both in a browser and IDE.

dtivel (changes requested):

dtivel:
@jackmtpt, thank you for your contribution!  We appreciate it.

Can you open an issue with repro steps and link it?  It would help us better understand the scenario you're trying to improve.

I left some feedback.  I also created a [branch](https://github.com/dotnet/sign/tree/dtivel/better-manifest-file-detection) with my suggested changes.  You can cherry pick the latest commit if you want.

dtivel:
There's been no activity here for weeks, so I'm closing this in favor of https://github.com/dotnet/sign/pull/773.

Issue #684, opened 2024-05-08
Starting with version 0.9.1-beta.24123.2 error on tempfile/folder


RonKoppelaar:
**Describe the bug**
Starting with version 0.9.1-beta.24123.2 I got file in use errors when using the sign tool. Its depending on machine type/configuration but unclear why. I have two build machines both run on W2019 

**Repro steps**
Run Sign.exe for one app file (Business Central) using certificate from Azure keyvault. 
Invoke the cmd from PS ISE.

```
        . $signingToolExe code azure-key-vault `
            --azure-key-vault-url "https://$KeyVaultName.vault.azure.net/" `
            --azure-key-vault-certificate $CertificateName `
            --azure-key-vault-client-id $ClientId `
            --azure-key-vault-client-secret (Get-SecretValue $ClientSecret) `
            --azure-key-vault-tenant-id $TenantId `
            --description $Description `
            --description-url $DescriptionUrl `
            --file-digest $DigestAlgorithm `
            --timestamp-digest $DigestAlgorithm `
            --timestamp-url $TimestampService `
            --verbosity $Verbosity `
            $FileToSign
```
Error stack:

VERBOSE: Running Invoke-cdsaALAppSign. PSBoundParameters: 
 KeyVaultName:CodeSignERP CertificateName:CodeSignCertificate FilesToSign:C:\temp\Zig_Zig365 Insights Connector_23.23.0.0.app Description:test signing DescriptionUrl:https:\\www.zig.nl Verbose:True Verbosity:Information 
VERBOSE: Start find sign tool
VERBOSE: Instal Sign tool
VERBOSE: Installing signing tool version 0.9.1-beta.24123.2 in C:\Users\BUILDA~1.JVF\AppData\Local\Temp\1\SigningTool-0.9.1-beta.24123.2
VERBOSE: Getting sign tool C:\Users\BUILDA~1.JVF\AppData\Local\Temp\1\SigningTool-0.9.1-beta.24123.2\Sign.exe
VERBOSE: Using sign tool: C:\Users\BUILDA~1.JVF\AppData\Local\Temp\1\SigningTool-0.9.1-beta.24123.2\Sign.exe
VERBOSE: Signing file:  C:\temp\Zig_Zig365 Insights Connector_23.23.0.0.app
info: Sign.Core.ISigner[0]
      Submitting C:\temp\Zig_Zig365 Insights Connector_23.23.0.0.app for signing.
info: Sign.Core.ISigner[0]
[...]

PR #762, opened 2024-09-17
Warn if MSVCRT 14 not detected

dtivel:
Address https://github.com/dotnet/sign/issues/684

This change warns if MSVCRT 14 is not installed and updates documentation prerequisites.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,23 @@
 
 This project aims to make it easier to integrate secure code signing into a CI pipeline by using cloud-based hardware security module(HSM)-protected keys. This project is part of the [.NET Foundation](https://www.dotnetfoundation.org/) and operates under their [code of conduct](https://www.dotnetfoundation.org/code-of-conduct). It is licensed under [MIT](https://opensource.org/licenses/MIT) (an OSI approved license).
 
+## Prerequisites
+
+- An up-to-date x64-based version of Windows currently in [mainstream support](https://learn.microsoft.com/lifecycle/products/)
+- [.NET 8 SDK or later](https://dotnet.microsoft.com/download)
+- [Microsoft Visual C++ 14 runtime](https://aka.ms/vs/17/release/vc_redist.x64.exe)
+
+## Install
+
+To install Sign CLI in the current directory, open a command prompt, path to the directory where Sign CLI will be installed, and execute:
+
+```
+dotnet new tool-manifest
+dotnet tool install --local --prerelease sign
+```
+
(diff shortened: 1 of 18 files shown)


src/Sign.Cli/Program.cs
```diff
@@ -44,6 +44,18 @@ internal static async Task<int> Main(string[] args)
...
                 AddEnvironmentPath(netfxDir);
 
+                string systemDirectoryPath = Environment.GetFolderPath(Environment.SpecialFolder.System);
+
+                // NavSip.dll has a dependency on this.
+                string vcRuntime140FilePath = Path.Combine(systemDirectoryPath, "vcruntime140.dll");
+
+                if (!File.Exists(vcRuntime140FilePath))
+                {
+                    Console.ForegroundColor = ConsoleColor.Yellow;
```
dlemstra, line 54: Maybe this could be an a separate method? `WriteWarning`?

dtivel: OK.  I _really_ wanted to use `ILogger.LogWarning(...)`, but I couldn't see how to get an `ILogger` this early in startup and honor logging options and reuse the same logger throughout.

README.md
```diff
@@ -4,6 +4,23 @@
...
+- An up-to-date x64-based version of Windows currently in [mainstream support](https://learn.microsoft.com/lifecycle/products/)
+- [.NET 8 SDK or later](https://dotnet.microsoft.com/download)
+- [Microsoft Visual C++ 14 runtime](https://aka.ms/vs/17/release/vc_redist.x64.exe)
+
+## Install
+
+To install Sign CLI in the current directory, open a command prompt, path to the directory where Sign CLI will be installed, and execute:
+
+```
+dotnet new tool-manifest
```
clairernovotny, line 18: The manifest shouldn't be created. That makes it need `dotnet sign`, which is incorrect as this is a standalone tool.

This command works and doesn't hardcode a version:
`dotnet tool install --tool-path . sign --prerelease`

Then it can be executed from the local directory with just `sign` (no dotnet needed)`

README.md
```diff
@@ -4,6 +4,23 @@
...
+dotnet tool install --local --prerelease sign
+```
+
+To run Sign CLI, execute `dotnet sign` from the same directory.
```
clairernovotny, line 22: this should just be `sign.exe` if installed with the `--tool-path .` parameter above.

clairernovotny (changes requested):

Issue #751, opened 2024-08-04
CI:  test failure details not logged


dtivel:
When a test fails in a public CI build, test failure details are not persisted.  The build log shows something like this, which is insufficient to determine which test failed or why it failed.

```
XUnit : error : Tests failed: D:\a\_work\1\s\artifacts\TestResults\Release\Sign.Core.Test_net8.0_x64.html [net8.0|x64] [D:\a\_work\1\s\test\Sign.Core.Test\Sign.Core.Test.csproj]
##[error]XUnit(0,0): error : (NETCORE_ENGINEERING_TELEMETRY=Test) Tests failed: D:\a\_work\1\s\artifacts\TestResults\Release\Sign.Core.Test_net8.0_x64.html [net8.0|x64]
```

Ideally, in the case of failure, the log would be attached as a build artifact.

PR #763, opened 2024-09-17
CI:  upload test results on failure

dtivel:
Resolve https://github.com/dotnet/sign/issues/751.

On failure, try to upload test results as an artifact.

### DIFF
--- a/.vsts-pr.yml
+++ b/.vsts-pr.yml
@@ -41,3 +41,12 @@ stages:
           name: Build
           displayName: Build and run tests
           condition: succeeded()
+        - task: PublishTestResults@2
+          displayName: 'Publish test results'
+          inputs:
+            testResultsFormat: xUnit
+            testResultsFiles: '$(Build.SourcesDirectory)/artifacts/TestResults/**/*.xml'
+            mergeTestResults: true
+            searchFolder: $(System.DefaultWorkingDirectory)
+            testRunTitle: sign unit tests - $(Agent.JobName)
+          condition: failed()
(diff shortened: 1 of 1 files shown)


.vsts-pr.yml
```diff
@@ -40,4 +40,13 @@ stages:
...
+          condition: succeeded()
+        - task: PublishTestResults@2
+          displayName: 'Publish test results'
+          inputs:
+            testResultsFormat: xUnit
+            testResultsFiles: '$(Build.SourcesDirectory)/artifacts/TestResults/**/*.xml'
+            mergeTestResults: true
+            searchFolder: $(System.DefaultWorkingDirectory)
+            testRunTitle: sign unit tests - $(Agent.JobName)
+          condition: failed()
```
joelverhagen, line 52: Curious why not use `always()` or `succeededOrFailed()`? Would you want test results on success case so you have a baseline to compare to?

dtivel: Maybe, but the problem this change solves is that a PR build sometimes fails on a flaky test, but we don't know which test or the failure reason.  With this change, you can see that information ([example](https://dev.azure.com/dnceng-public/public/_build/results?buildId=811149&view=ms.vss-test-web.build-test-results-tab)) and work on test flakiness.

What information do you want from the baseline success case?  The tests don't log diagnostic information (e.g.:  `ITestOutputHelper`), so the baseline information they offer now is that they passed and execution time.

[The internal pipeline already uploads test results in failure or success.](https://github.com/dotnet/sign/blob/main/.vsts-ci.yml#L129)

joelverhagen: Ah, I didn't know about the other internal pipeline. 

Might be nice to be able to confirm a new test actual ran (and wasn't ignored by the test runner) and it would be consistent with the other pipeline. But I don't feel strongly. Feel free to ignore.

dtivel: I changed it to `succeededOrFailed()`.  Thanks for the feedback.

joelverhagen (approved):

Issue #771, opened 2024-10-10
Faulty resource reuse in CertificateStoreService.cs


MeikTranel:
**Describe the bug**

When using the local (windows) certificate store to find a cert the `CertificateStoreService` prints that its searchin in Azure Key Vault, because the log message was likely copied from the azure key vault provider. See: https://github.com/dotnet/sign/blob/ac9e8b416cd6e8692f1eea1ac6678d7e5c464216/src/Sign.SignatureProviders.CertificateStore/CertificateStoreService.cs#L118

Uses the same resource key as:
https://github.com/dotnet/sign/blob/ac9e8b416cd6e8692f1eea1ac6678d7e5c464216/src/Sign.SignatureProviders.KeyVault/KeyVaultService.cs#L79

which is:
https://github.com/dotnet/sign/blob/ac9e8b416cd6e8692f1eea1ac6678d7e5c464216/src/Sign.Core/Resources.resx#L190-L192

Not sure of the semantics of TrustedSigningService but that service is also reusing that resource key:
https://github.com/dotnet/sign/blob/ac9e8b416cd6e8692f1eea1ac6678d7e5c464216/src/Sign.SignatureProviders.TrustedSigning/TrustedSigningService.cs#L70

![Image](https://github.com/user-attachments/assets/8a565092-8abd-4604-9bfb-ef71114ae5e8)

**Expected behavior**
Verbose logs should print that it was searching in the windows cert store. Bonus points if that log also provides where exactly it searched and what it found "Issue To" or the SHA1 Hash or something would be helpful.

**Actual behavior**
A clear and concise description of what actually happened.

**Additional context**

```
 λ dotnet sign --version                                                                                                                                                       
0.9.1-beta.24469.1+5516908609085978174904945d54144c117041dc
```

```
 λ dotnet --info
.NET SDK:
 Version:           9.0.100-rc.1.24452.12
 Commit:            81a714c6d3
 Workload version:  9.0.100-manifests.67cd1eb6
 MSBuild version:   17.12.0-preview-24422-09+d17ec720d

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22631
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\9.0.100-rc.1.24452.12\

.NET workloads installed:
Configured to use loose manifests when installing new manifests.
 [aspire]
   Installation Source: VS 17.11.35327.3, VS 17.12.35323.107
[...]

PR #772, opened 2024-10-10
Moved resources to signature provider libraries.

dlemstra:
This PR moves some of the resources from the core library to the signature provider libraries. And I also made some improvements to the logging messages in the CertificateStoreService that were suggested in #771.


dtivel (approved):
LGTM, @dlemstra.  Thank you!

Issue #681, opened 2024-04-05
Sign many applications and manifestfiles in batch for click-once


klamfeldt:
Is your feature request related to a problem? Please describe.
Cant sign many application-files in one batch. We have a setup where we have a *.application and *.manifest file per customer and we have about 100 customers. I would like to sign all files in one run.

Describe the solution you'd like
When signing CustomerA.application file the program should look for a manifestfile named CustomerA.dll.manifest

Describe alternatives you've considered
My workaround now is to tempory move customerA.application and customerA.dll.manifest to a temporary folder and sign them and them move them back.

jackmtpt:
Do you actually have a .manifest per customer, or just per-version of your application? We do something similar - with a .manifest for e.g. v1 of the app and another for v2, and then per-customer .application files with a unique update URL and that we configure to point to either v1 or v2 of the .manifest file.

I merged some changes to this tool a few weeks ago that should support signing multiple .application files but there's not been a release since then so you'd have to compile it yourself.

klamfeldt:
Yes we have 1 manifestfile for each customer, since we have an unique URL for each customer. I have compiled the latest code and run but fails with "Sequence contains more than one matching element" from  \ClickOnceSignatureProvider.cs:line 130
which has a SingleOrDefault
FileInfo? manifestFile = filteredFiles.SingleOrDefault(f => ".manifest".Equals(f.Extension, StringComparison.OrdinalIgnoreCase));

ottD:
We are facing the same issue as @klamfeldt outlined as our solution contains multiple .manifest files in the same directory. The current implementation of the ClickOnceSigner assumes a single .manifest file which is pretty limiting and renders the sign tool unusable without having to implement some sort of workaround.

jackmtpt:
I think I can put together a fix for this.

jackmtpt:
I've written #758 which should hopefully solve your problem. I've given it a test locally with a clickonce package that contains a the proper clickonce app manifest (MyApp.dll.manifest) as well as a separate .dll.manifest which isn't related to clickonce at all; it's just a regular windows manifest file and it worked fine.

For anyone with an app manifest (.dll.manifest/.exe.manifest) per customer though, I am curious why you need that because the URL isn't contained within those files - their contents depends only on the hashes of the binaries, so they should be 1:1 with versions of your application so unless you're actually building your app separately for each customer then you only need one app manifest per version, and you can construct however many deployment manfiests (.application) you need with different update URLs.

[...]

PR #773, opened 2024-10-11
ClickOnce:  improve application manifest file detection

dtivel:
Resolve https://github.com/dotnet/sign/issues/681

This PR replaces https://github.com/dotnet/sign/pull/758.

CC @clairernovotny, @javierdlg, @jackmtpt

### DIFF
--- a/src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
+++ b/src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
@@ -22,6 +22,7 @@ namespace Sign.Core
         private readonly IManifestSigner _manifestSigner;
         private readonly ParallelOptions _parallelOptions = new() { MaxDegreeOfParallelism = 4 };
         private readonly IFileMatcher _fileMatcher;
+        private readonly IXmlDocumentLoader _xmlDocumentLoader;
 
         // Dependency injection requires a public constructor.
         public ClickOnceSigner(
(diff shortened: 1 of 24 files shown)


src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs
```diff
@@ -273,5 +281,60 @@ public void CopySigningDependencies(FileInfo deploymentManifestFile, DirectoryIn
...
+                Logger.LogDebug(Resources.ApplicationManifestNotFound);
+                return false;
+            }
+
+            XmlNode? node = dependentAssemblies.Item(0);
+            if (node is null || node.Attributes is null)
+            {
+                Logger.LogDebug(Resources.ApplicationManifestNotFound);
+                return false;
+            }
```
dtivel, line 314: There seems to be a bug here.  The comment asserts that there should only be one dependentAssembly element.  However, it seems that there can be multiple.

From https://learn.microsoft.com/visualstudio/deployment/dependency-element-clickonce-deployment?view=vs-2022#elements-and-attributes:

> The `dependency` element is required. It has no attributes. A deployment manifest can have multiple `dependency` elements.

This means this change would break users with multiple `dependency` elements.  This method needs smarter application manifest identification.

src/Sign.Core/Xml/IXmlDocumentLoader.cs
```diff
@@ -0,0 +1,13 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+using System.Xml;
+
+namespace Sign.Core
+{
+    internal interface IXmlDocumentLoader
+    {
+        XmlDocument Load(FileInfo file);
```
dlemstra, line 11: Wondering why you picked `XmlDocument` instead of `XDocument` here.

andymozza:
Can confirm this resolved the issue for us with two VSTO applications where manifest signing was failing.

Issue #769, opened 2024-10-06
Bug: Incomplete PowerShell module signing support


MichaelGrafnetter:
**Describe the bug**
Although the tool can sign PowerShell script files (`*.ps1` and `*.psm1`) with Azure Trusted Signing, support for additional extensions used by PowerShell (`*.psd1` and `*.ps1xml`) seems to be missing. As a result, PowerShell modules containing manifests and format definitions cannot be digitally signed by this tool.

**Repro steps**
```bat
sign code trusted-signing
        **/*.ps*1
        --trusted-signing-endpoint "${{ secrets.SIGNING_ENDPOINT }}"
        --trusted-signing-account "${{ secrets.SIGNING_ACCOUNT_NAME }}"
        --trusted-signing-certificate-profile "${{ secrets.SIGNING_PROFILE_NAME }}"
        --file-digest SHA256
        --timestamp-digest SHA256
        --timestamp-url "http://timestamp.acs.microsoft.com"
        --verbosity Information
```

**Expected behavior**
All `*.psm1` and `*.psd1` files should get digitally signed.

**Actual behavior**
Only `*.psm1` files get signed, while `*.psd1` files are ignored.

**Suggested resolution**
The issue might simply be resolved by adding `*.psd1` and `*.ps1xml` to the list of supported extensions in the [AzureSignToolSigner.cs](https://github.com/dotnet/sign/blob/04303f6df1622299042fb28f6060fd073ccd2dd6/src/Sign.Core/DataFormatSigners/AzureSignToolSigner.cs) file.

**Additional context**
Sign version 0.9.1-beta.24469.1+5516908609085978174904945d54144c117041dc

dtivel:
@MichaelGrafnetter, https://www.nuget.org/packages/sign/0.9.1-beta.24529.1 has the fix.

Signing .xaml files is currently unsupported.  https://github.com/dotnet/sign/issues/778

MichaelGrafnetter:
Thanks @dtivel, good work! I completely missed the remaining 2 extensions used by Powershell.

PR #777, opened 2024-10-26
PowerShell:  add support for signing .ps1xml and .psd1 files

dtivel:
Resolve https://github.com/dotnet/sign/issues/769

This change adds support for signing .ps1xml, .psd1, and .cdxml files.

* Extracted app initialization code in Sign.Cli/Program.cs into a reusable class `AppInitializer`.
* Refactored test certificate creation code into `SelfIssuedCertificateCreator`.
* Refactored test asset reading code into `TestAssets`.
* Moved `TemporaryFile` into Sign.TestInfrastructure project.
* Created `RequiresElevationTheory` attribute and `TrustedCertificateFixture` for tests that require trusted certificates.  These tests will run when elevated or when running in CI and skip when not.
* Added tests for PowerShell file signing.


clairernovotny (approved):

test/Sign.TestInfrastructure/ElevatedTheoryAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Sign.TestInfrastructure
+{
+    public sealed class ElevatedTheoryAttribute : TheoryAttribute
+    {
+        private const string SkipReason = "Test skipped because it requires elevation.";
+
+        private static readonly Lazy<bool> LazyIsElevated = new(IsElevated);
+
+        public override string Skip
```
joelverhagen, line 15: I'm curious about this approach. If for some reason the CI/CD doesn't run as elevated we'll miss the test coverage in an invisible way, right? I wonder if we should make this default to `SKIP = true` only when we're not running on CI/CD. I understand skipping automatically when running on devbox is nice. It would be confusing to contributors to run `dotnet test` on a git clone and immediately have failing tests. But we should never let these tests be skipped on CI/CD right?

One option I have used in the past is to detect our CI agent environment variable to do this "Is this CI" check. You could consider one of the variables here:
https://github.com/dotnet/reproducible-builds/blob/714df2e17493c71d0db054727f62a82810eeac7d/src/DotNet.ReproducibleBuilds/DotNet.ReproducibleBuilds.props#L15-L38

src/Sign.Core/DataFormatSigners/AzureSignToolSigner.cs
```diff
@@ -57,6 +57,8 @@ public AzureSignToolSigner(
                 ".mst",
                 ".ocx",
                 ".ps1",
+                ".ps1xml",
```
joelverhagen, line 60: I tried to understand the full list of extensions that could be passed in here, to see if there are any other missing ones (e.g. other PowerShell-ish extensions). It seems like the only allow-list that AzureSignTool has is adding a SIP extension for Appx extensions. Do we have a way of knowing all the extensions that we possibly could provide here? 

Aside from testing it, how do we actually know ps1xml and psd1 are supported by the underlying native signing methods?

dtivel: In the general case, I don't know.  For PowerShell, there are [6 file extensions which support signing](https://github.com/PowerShell/PowerShell/blob/2f4f585e7fe075f5c1669397ae738c554fa18391/src/System.Management.Automation/security/SecurityManager.cs#L97-L106).  I updated this PR to support .cdxml.  I can't add support for .xaml because of https://github.com/dotnet/sign/issues/778.

test/Sign.TestInfrastructure/ElevatedTheoryAttribute.cs
dtivel: Good feedback.  I updated it to execute if running in CI or elevated.

joelverhagen (approved):

PR #806, opened 2024-11-19
Add action to auto close inactive issues as per our triage policy

clairernovotny:
Add action to auto close inactive issues as per our triage policy

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,19 @@
+name: 'Close stale issues and PRs'
+
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 7 days with no activity after asking for more info. Comment or this will be closed in 4 days.'
+          close-issue-message: 'This issue was closed because it has been stalled for 10 days with no activity. This can be reopened if additional information is provided.'
+          days-before-issue-stale: 10          
+          days-before-issue-close: 4
+          days-before-pr-stale: -1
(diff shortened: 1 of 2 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
+name: 'Close stale issues and PRs'
```
dtivel, line 1: I don't think we should touch PR's yet.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+
+permissions:
+  issues: write
+  pull-requests: write
```
dtivel, line 5: I don't think we should touch PR's yet.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 7 days with no activity after asking for more info. Comment or this will be closed in 4 days.'
```
dtivel, line 17: Triage policy says 14 days not (7 + 4) 11 days.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+          close-issue-message: 'This issue was closed because it has been stalled for 10 days with no activity. This can be reopened if additional information is provided.'
```
dtivel, line 18: Not sure where 10 came from.  It should be 14 days.

.github/workflows/stale.yml
clairernovotny: Fixed the timeouts -- the intent is to warn at 10 days and close 4 days later.

dtivel (approved):

PR #811, opened 2025-02-10
Update dependencies

dtivel:
Resolve https://github.com/dotnet/sign/issues/810.

This change updates dependencies used to build, run, and package Sign CLI.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,9 +1,9 @@
 {
     "tools": {
-        "dotnet": "9.0.100",
+        "dotnet": "9.0.102",
         "runtimes": {
             "dotnet/x64": [
-                "8.0.11"
+                "9.0.1"
             ]
         },
         "xcopy-msbuild": "17.10.0-pre.4.0"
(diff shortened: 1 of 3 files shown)


erdembayar (approved):

global.json
```diff
@@ -1,9 +1,9 @@
 {
     "tools": {
-        "dotnet": "9.0.100",
+        "dotnet": "9.0.102",
         "runtimes": {
             "dotnet/x64": [
-                "8.0.11"
```
joelverhagen, line 6: I see the project still targets `net8.0`. At what cadence do we move the `tools` and `sdk` sections to new runtime or SDK versions? I'm confused on how all of these relate for Sign CLI and the desired development environment.

joelverhagen (approved):

global.json
dtivel: Good point.  Because Sign CLI is a tool, `tools` affect Sign CLI itself when run from this directory.  I've updated it to use 8.x and added a comment.

SDK versions can be later versions.  We've aimed to use the latest .NET release.

dtivel: Ah, now I remember.  `tools.dotnet` and `tools.runtimes` are out of sync because Sign CLI depends on the .NET 8 runtime, but the .NET 8 version of `xcopy-msbuild` has CG alerts, which are only resolved by updating to .NET 9 SDK for tools.  See https://github.com/dotnet/msbuild/commit/977f965b0e18488a0768da26277177524f6d1107

joelverhagen (approved):

PR #816, opened 2025-02-13
Allow 'code' commands to accept multiple file arguments

### LINKED ISSUE #815: "trusted-signing <file(s)>" doesn't accept multiple files
**Describe the bug**

Run `sign code trusted-signing --help`. It prints:

```
Usage:
  sign code trusted-signing <file(s)> [options]
```

The expectation is that one can pass multiple absolute paths (to different DLLs or EXEs) to the command. However, if you do so, you get the error:

```
Unrecognized command or argument 'C:\Path\To\Dll2.dll'.
```

**Repro steps**

Run:

```
sign code trusted-signing C:\Path\To\Dll1.dll C:\Path\To\Dll2.dll -tse https://wus2.codesigning.azure.net/ -tsa REDACTED -tscp REDACTED -fd sha256 -t http://timestamp.acs.microsoft.com -td sha256 -v Debug
```

You get the error:

```
Unrecognized command or argument 'C:\Path\To\Dll2.dll'.
```

**Expected behavior**

Multiple files can be specified and all will be signed.

**Actual behavior**

The second file generates an error.

**Additional context**
- Include the output of `sign --version`: 0.9.1-beta.24529.1+029731ca2df9d41933a09a1cc9212595a7c7932a
- Include the output of `dotnet --info`.

```
.NET SDK:
 Version:           9.0.102
 Commit:            cb83cd4923
 Workload version:  9.0.100-manifests.c4f6226a
 MSBuild version:   17.12.18+ed8c6aec5
```
 
- Add any other context about the problem here.

Appears to be related to this: https://github.com/dotnet/sign/blob/4dcffd3321562f8e3a9de861c1f226ac506da0f3/src/Sign.Cli/TrustedSigningCommand.cs#L21

The simplest fix would be to change the documentation to `<file>`, not `<file(s)>`. However, the desired behavior would be for multiple files to be supported.

I was going to try using a file list as a workaround but I saw https://github.com/dotnet/sign/issues/768 which indicates that that might not work.

bgrainger:
Fixes #815

The command documentation specifies that these commands take a "file(s)" parameter. They currently accept only a single argument, although it can be a glob to specify multiple files. This changes them to accept multiple filenames as separate arguments, similar to the existing behavior of signtool and azuresigntool.

### DIFF
--- a/src/Sign.Cli/AzureKeyVaultCommand.cs
+++ b/src/Sign.Cli/AzureKeyVaultCommand.cs
@@ -22,7 +22,7 @@ namespace Sign.Cli
         internal Option<string> CertificateOption { get; } = new(["--azure-key-vault-certificate", "-kvc"], AzureKeyVaultResources.CertificateOptionDescription);
         internal AzureCredentialOptions AzureCredentialOptions { get; } = new();
 
-        internal Argument<string?> FileArgument { get; } = new("file(s)", Resources.FilesArgumentDescription);
+        internal Argument<List<string>?> FilesArgument { get; } = new("file(s)", Resources.FilesArgumentDescription) { Arity = ArgumentArity.OneOrMore };
 
         internal AzureKeyVaultCommand(CodeCommand codeCommand, IServiceProviderFactory serviceProviderFactory)
             : base("azure-key-vault", AzureKeyVaultResources.CommandDescription)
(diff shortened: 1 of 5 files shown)


dtivel:
@bgrainger, this change looks good.  Can you please rebase on latest and update?

dtivel (approved):

jeffkl (approved):

dtivel:
@bgrainger, this should be available in [0.9.1-beta.25224.2](https://www.nuget.org/packages/sign/0.9.1-beta.25224.2).

PR #834, opened 2025-03-20
Silence CA2255 (The ModuleInitializer attribute should not be used in libraries) through the .editorconfig.

dlemstra:
While building the library today I suddenly got a build error due to CA2255 that mentions that we should not be using the ModuleInitializer attribute. This small patch updates the `.editorconfig` to silence that warning/error.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,6 +18,9 @@ csharp_prefer_simple_using_statement = false
 # See https://github.com/dotnet/roslyn-analyzers/issues/5626
 dotnet_diagnostic.CA2254.severity = none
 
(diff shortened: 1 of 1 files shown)


dtivel:
@dlemstra, thanks for the PR.  Out of curiosity, what code is raising CA2255?  When I build clean, I get no warnings.  (How can I reproduce the CA2255?)

dlemstra:
In one of the unit test we use that attribute. Maybe it depends on the Visual Studio version?

dtivel (approved):

erdembayar (approved):

Issue #837, opened 2025-03-24
Wrong timestamp with English (Denmark) regional settings when signing a .VSIX


ErikEJ:
**Describe the bug**

Using "Recommended" which was "English (Denmark)"

Certificate is invalid, as the format of the signaturetime does not match the value

`<SignatureTime xmlns="http://schemas.openxmlformats.org/package/2006/digital-signature"><Format>YYYY-MM-DDThh:mm:ss.sTZD</Format><Value>2025-03-24T10.56.12.2-07:00</Value></SignatureTime>
`
**Repro steps**

`sign code certificate-store -cfp ABC... -k BEEF -csp "SimplySign CSP" -t http://time.certum.pl "EF Core Power Pack v2.6.959.vsix"
`

**Additional context**

as you can see I am signing a .vsix file. I had my regional settings set to English (Denmark). Switching to Danish (Denmark) fixes the issue, and produces a valid cert

`<SignatureTime xmlns="http://schemas.openxmlformats.org/package/2006/digital-signature"><Format>YYYY-MM-DDThh:mm:ss.sTZD</Format><Value>2025-03-24T11:20:22.5-07:00</Value></SignatureTime>`

- Include the output of `sign --version`. : 0.9.1-beta.25169.2+c14370f70687ab46f23e7dbcaf1a401ca2a0c494
- Include the output of `dotnet --info`.

```
.NET SDK:
 Version:           9.0.201
 Commit:            071aaccdc2
 Workload version:  9.0.200-manifests.a3a1a094
 MSBuild version:   17.13.13+1c2026462

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.26100
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\9.0.201\

.NET workloads installed:
There are no installed workloads to display.
Configured to use loose manifests when installing new manifests.

Host:
  Version:      9.0.3
  Architecture: x64
  Commit:       831d23e561

.NET SDKs installed:
  8.0.407 [C:\Program Files\dotnet\sdk]
  9.0.201 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.14 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 9.0.3 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

PR #839, opened 2025-03-28
VSIX:  fix signature timestamp format string bug

dtivel:
Resolve https://github.com/dotnet/sign/issues/837.

By default, `DateTimeOffset.ToString(string)` uses the current culture.  For en-DK, the time separator is `.` not `:`.  So, the result is a value which doesn't follow the [OPC signature timestamp format](https://learn.microsoft.com/windows/win32/api/msopc/ne-msopc-opc_signature_time_format).

This change forces the `DateTimeOffset.ToString(...)` call to use `CultureInfo.InvariantCulture`.  This produces the expected result regardless of the current culture.

### DIFF
--- a/test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageSigningTests.cs
+++ b/test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageSigningTests.cs
@@ -2,8 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE.txt file in the project root for more information.
 
+using System.Globalization;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
+using System.Xml;
 using Sign.Core.Timestamp;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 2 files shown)


ErikEJ (commented):
LGTM

joelverhagen (approved):

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageSigningTests.cs
```diff
@@ -189,6 +191,73 @@ public void ShouldRemoveSignature()
...
+            try
+            {
+                // This test only works if the current culture is one of a set of cultures that includes en-DK that
+                // that repro the original bug.  However, because tests should not rely on a specific culture being
+                // installed, we'll create a custom culture just for this test.
+                var customCulture = (CultureInfo)CultureInfo.InvariantCulture.Clone();
+
+                customCulture.DateTimeFormat.TimeSeparator = ".";
+
+                CultureInfo.CurrentCulture = customCulture;
```
joelverhagen, line 208: Will this cause problems with other tests running in parallel? 

I see the docs say:
> The culture used by the current thread and task-based asynchronous operations.

But it's not clear to me that the async context is properly isolated in xunit.

dtivel: Good question.

In an async context, the calling thread determines the culture for subsequent thread continuations.  See [this doc section](https://learn.microsoft.com/dotnet/fundamentals/runtime-libraries/system-globalization-cultureinfo-currentculture#how-a-threads-culture-is-determined) and [this comment](https://github.com/dotnet/runtime/issues/48077#issuecomment-776231121).

Also, [these tests run sequentially.](https://github.com/dotnet/sign/blob/808759445b7eb95242aa1a67cdaa51026ce3d04d/test/Sign.Core.Test/TestInfrastructure/Server/CertificatesFixture.cs#L9)

Also, [this test is synchronous not async.](https://github.com/dotnet/sign/blob/808759445b7eb95242aa1a67cdaa51026ce3d04d/test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageSigningTests.cs#L195)

Other tests won't be affected by this.

erdembayar (approved):

Issue #842, opened 2025-03-31
ClickOnce:  propose new signing algorithm


dtivel:
After working on https://github.com/dotnet/sign/pull/773 for some time, it's clear that Sign CLI's ClickOnce signing algorithm needs improvement.

PR #843, opened 2025-03-31
ClickOnce:  add ClickOnce signing algorithm spec

dtivel:
See https://github.com/dotnet/sign/issues/842.

This PR adds a specification for changes to the ClickOnce signing algorithm.  The goal is to make ClickOnce signing simpler, more correct, more efficient, and easier to maintain over time.

Rendered view:  https://github.com/dotnet/sign/blob/cee18cb08d824564dd087974d00d11dde0fa2972/docs/specs/ClickOnce-Signing-Algorithm.md

(The rendered view is best viewed in light mode, because the animated GIF, which you can see clearly [here](https://github.com/dotnet/sign/pull/843/files?short_path=4b4a06c#diff-904311e17dc44ed412eac95dd16816a9b22f18235425cb0115f95dcbc701d0d4), does not display well in dark mode.)


docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,134 @@
...
+   * Note:  it seems like calling `Manifest.ResolveFiles()` would resolve the full file path for every file dependency in the application manifest.  However, this fails because `ResolveFiles()` assumes dependency files do not have `.deploy` file extension, but they do.  We could temporarily remove the `.deploy` extension and then call `ResolveFiles()` but a user's glob patterns might filter out files based on the `.deploy` extension.  The safest option is to resolve file paths ourselves.
+1. Copy all files from the previous step, including the application manifest file itself, to a temporary directory.
+1. Sign files in the following order:  files alongside the application manifest, the application manifest itself, then the deployment manifest.
+1. Copy the files back.
+1. If the signed deployment manifest file is a `.vsto` file, copy it to the versioned application manifest file directory and overwrite if necessary.
```
dtivel, line 133: I'm debating this.  I feel like Sign CLI should not do this; Sign CLI should focus on signing, not emulating behaviors of a build-time task.  Users who desire this behavior --- not just for `.vsto` files, but any ClickOnce app --- can do the copy after signing.

In trying to provide an equivalent experience to the VSTO build-time task, I think this will set up Sign CLI to do more non-signing-related chores.  This is not a sustainable direction for Sign CLI.

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,134 @@
...
+      myAddin.Excel.vsto
+
+      myAddin.Word.dll.manifest
+      myAddin.PowerPoint.dll.manifest
+      myAddin.Excel.dll.manifest
+  ```
+
+## Appendix B:  Proposed algorithm
+
+1. If a file has a `.vsto` or `.application` file extension, read it as a deployment manifest using [`ManifestReader.ReadManifest(...)`](https://learn.microsoft.com/dotnet/api/microsoft.build.tasks.deployment.manifestutilities.manifestreader.readmanifest?view=msbuild-17-netcore#microsoft-build-tasks-deployment-manifestutilities-manifestreader-readmanifest(system-io-stream-system-boolean)).  If file reading fails or the returned [`Manifest`](https://learn.microsoft.com/dotnet/api/microsoft.build.tasks.deployment.manifestutilities.manifest?view=msbuild-17-netcore) instance is not a [`DeployManifest`](https://learn.microsoft.com/dotnet/api/microsoft.build.tasks.deployment.manifestutilities.deploymanifest?view=msbuild-17-netcore), defer to next signer.
```
Tanya-Solyanik, line 118: What is the `next signer`?

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,134 @@
...
+1. Read the application manifest file using `ManifestReader.ReadManifest(...)`.
+1. Set `Manifest.ReadOnly` to `true` to ensure read-only mode.
+1. Interate through both [`AssemblyReferences`](https://learn.microsoft.com/dotnet/api/microsoft.build.tasks.deployment.manifestutilities.manifest.assemblyreferences?view=msbuild-17-netcore) and [`FileReferences`](https://learn.microsoft.com/dotnet/api/microsoft.build.tasks.deployment.manifestutilities.manifest.filereferences?view=msbuild-17-netcore), manually resolve `TargetPath` property to the base path of the application manifest file.
+   * Note:  it seems like calling `Manifest.ResolveFiles()` would resolve the full file path for every file dependency in the application manifest.  However, this fails because `ResolveFiles()` assumes dependency files do not have `.deploy` file extension, but they do.  We could temporarily remove the `.deploy` extension and then call `ResolveFiles()` but a user's glob patterns might filter out files based on the `.deploy` extension.  The safest option is to resolve file paths ourselves.
```
Tanya-Solyanik, line 129: Note that the user can publish assemblies with no .deploy extension as well.

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,134 @@
...
+1. Copy all files from the previous step, including the application manifest file itself, to a temporary directory.
```
Tanya-Solyanik, line 130: Why is this step needed?

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,134 @@
...
+1. Sign files in the following order:  files alongside the application manifest, the application manifest itself, then the deployment manifest.
```
Tanya-Solyanik, line 131: You should sign setup.exe too, unless you want to use a different key for it.

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Sign CLI signs "inside out".  For example, given a ClickOnce application, sign the assemblies inside, then sign the application.  Sign CLI aggregates different signers for formats like Authenticode, VSIX, NuGet, ClickOnce, etc.  Each signer is given a chance to sign a file.  Currently, the ClickOnce signer will try to sign if a file's extension is either [`.vsto` or `.application`](https://github.com/dotnet/sign/blob/5beaa81e1f786e610a2284641a41e47664e2d9c8/src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs#L58).  Internally, the ClickOnce signer will try to sign all the files it knows are related to ClickOnce signing.

Here "defer to next signer" means that the ClickOnce signer will give up and let the next signer have a try, but there are no other signers that handle `.vsto` or `.application` files.

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Sign CLI does not sign in place.  It copies files to a temporary location and copies back only if signing succeeded.  This can happen multiple times if it encounters an archive file.  Sign CLI extracts the archive to a temporary location, signs files, recreates the archive, and copies the archive back to the original location.

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: @Tanya-Solyanik, can you give more details on this?  Why have files with or without the `.deploy` extension?  (Why have the `.deploy` extension at all?)

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Good catch!  I covered it [there](https://github.com/dotnet/sign/pull/843/files#diff-4b4a06c79e12d36a20f2eba719fd1d56a9634b06d678fe75561f05130209b48eR49) but not here.

docs/specs/ClickOnce-Signing-Algorithm.md
Tanya-Solyanik: When application deploy is deployed to a web server, .exe and .dll extentions might not be downloadable as they might be blocked by the admin

docs/specs/ClickOnce-Signing-Algorithm.md
sujitnayak: Since it's not called out, I wanted to mention that if you are taking an existing published application and signing it's manifest/binaries, the deployment manifest will have to be re-generated after signing the application manifest to update its assembly identity in the deployment manifest.

You can see it happening for ClickOnce apps here:
https://github.com/dotnet/msbuild/blob/014f3faab9cc9a5c3387d04190fbca51abac34e7/src/Tasks/Microsoft.Common.CurrentVersion.targets#L6336

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,134 @@
...
+   * Note:  it seems like calling `Manifest.ResolveFiles()` would resolve the full file path for every file dependency in the application manifest.  However, this fails because `ResolveFiles()` assumes dependency files do not have `.deploy` file extension, but they do.  We could temporarily remove the `.deploy` extension and then call `ResolveFiles()` but a user's glob patterns might filter out files based on the `.deploy` extension.  The safest option is to resolve file paths ourselves.
+1. Copy all files from the previous step, including the application manifest file itself, to a temporary directory.
+1. Sign files in the following order:  files alongside the application manifest, the application manifest itself, the deployment manifest, then `setup.exe`.
```
Tanya-Solyanik, line 131: You can sign files programmatically using https://github.com/dotnet/msbuild/blob/a9d68ab58eab3c25b6378a90be8060dd1429a6ef/src/Tasks/ManifestUtil/SecurityUtil.cs#L552

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,251 @@
...
+* `--no-update-clickonce-manifest`: When specified, Sign CLI will sign manifest files without calling `ResolveFiles()` and `UpdateFileInfo()`. This is useful when re-signing a manifest whose dependencies have not changed.
+
+These options are mutually exclusive (see [Option interactions](#option-interactions)). Without these options, Sign CLI will discover, update, and sign the complete ClickOnce application (deployment manifest, application manifest, and all referenced payload files).
+
+**Note**: Both `--no-sign-clickonce-deps` and `--no-update-clickonce-manifest` are only valid when used with `--use-new-clickonce-signing`. Attempting to use these options without enabling the new ClickOnce signing behavior will result in an error.
+
+### Rollout strategy
+
+The new ClickOnce signing algorithm is **opt-in**. Users must pass the `--use-new-clickonce-signing` flag to enable the proposed behavior. Without this flag, Sign CLI continues to use the current algorithm described in Appendix A.
```
jeffkl: Would it instead make sense to have a "version" of clickonce signing. The default today would be version 1 and this new one would be version 2.

The command-line argument would something like `--signing-version 1` and `--signing-version 2`.

Then we could consider logging a warning when people are using signing version 1 saying that version 2 is better.  Eventually we could make version 2 the default, and then users could still specify version 1 as an escape hatch in the future.

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,251 @@
...
+1. Sometimes [manifests need to be re-signed](https://learn.microsoft.com/visualstudio/deployment/how-to-re-sign-application-and-deployment-manifests?view=vs-2022). For re-signing, users need to be able to disable implicit signing of related files. For example, a user should be able to re-sign only a deployment manifest or just deployment and application manifests without re-signing payload files.
+
+## Proposed solution
+
+Given a deployment manifest file as a starting point, the algorithm will be updated to:
+
+1. Load the deployment manifest, locate the referenced application manifest, and refuse to continue if it is missing.
+1. Stage only the files referenced by the manifests, sign the payloads first, then the application manifest, then the deployment manifest, and finally the bootstrapper.
+1. After each signing stage, refresh manifest metadata so hashes, sizes, and entry-point information are consistent with the newly signed bits.
```
jeffkl: Curious is there are plans to make the whole operation atomic where if any one task fails, none of the files are signed or if signing proceeds until one failure occurs, leaving some files signed and other unsigned.

docs/specs/ClickOnce-Signing-Algorithm.md
kartheekp-ms: Given that we are still in the pre-release phase for SignCLI, I think breaking changes might be okay until the major release. I am proposing new behavior by default have a different option or an environment variable to fall back to the current behavior.

dtivel: @jeffkl, would it be generic (i.e., `--signing-version`) or ClickOnce-specific (e.g., `--clickonce-signing-version`)?  It seems like it should be the latter to enable signing parts to move independently of one another.

jeffkl: Yes it should be specific to clickonce. I originally thought clickonce signing was its own gesture but its not

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: After taking another look at the code, I confirmed that Sign CLI does not currently sign files atomically. If some files in a batch succeed and others fail, the succeeded files are already overwritten with their signed versions while the failed files remain untouched. This spec also does not intend to address that, even for ClickOnce signing. Atomicity is orthogonal to the core ClickOnce signing issues this spec covers. That said, it's a great callout. Do you want to open a separate issue for it?

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: I think [@jeffkl's proposal](https://github.com/dotnet/sign/pull/843#discussion_r3119083303) about introducing a v1/v2 choice strikes a good balance.  It gets the new behavior out there quickly while giving us room to tune it based on real-world feedback before making any disruptive default changes. Users who depend on v1 behavior aren't broken in the meantime.

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Thanks, @sujitnayak. The algorithm already had the required ordering, and I clarified that after signing the application manifest, Sign CLI re-resolves and updates the deployment manifest's hash, size, and entry-point identity before signing it: [updated algorithm](https://github.com/dotnet/sign/blob/dtivel/clickonce-spec/docs/specs/ClickOnce-Signing-Algorithm.md#L242-L242).

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Thanks, @jeffkl. I adopted a ClickOnce-specific `--clickonce-signing-version <1|2>` selector. Version 1 remains the initial default, version 2 is opt-in, and the rollout allows a later warning and default change while retaining version 1 as an escape hatch: [versioning and rollout](https://github.com/dotnet/sign/blob/dtivel/clickonce-spec/docs/specs/ClickOnce-Signing-Algorithm.md?plain=1#L82-L106).

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,290 @@
...
+1. Before signing begins, temporarily rename staged files whose names end with `.deploy` to their base names (for example, `MyApp.dll.deploy` → `MyApp.dll`).
+1. Discover applicable adjacent executables by checking if `setup.exe` or `Launcher.exe` exists in the same directory as the deployment manifest. `setup.exe` is an optional prerequisite bootstrapper; `Launcher.exe` launches the .NET application but does not participate in ClickOnce activation. A launcher or bootstrapper in a different directory or with a different name is not implicitly discovered as a dependency, but it will still be signed through standard Authenticode signing if matched by the user's file patterns.
+1. Sign payload files.
+1. Call `ApplicationManifest.UpdateFileInfo()` to refresh payload hashes, sizes, and identities, then restore the `.deploy` suffixes. (`UpdateFileInfo()` hashes files at their `ResolvedPath`, which does not include `.deploy`; the suffixes must be absent when hashes are computed.) Make the signed payload results available to waiting callers and complete their coordinated signing operations. Then sign the application manifest, make its signed result available, and complete its operation.
```
kartheekp-ms: `UpdateFileInfo()` (the parameterless overload) computes **SHA-1** hashes, not SHA-256.

In `Util.GetFileInfoImpl` ([MSBuild `ManifestUtil/Util.cs`](https://github.com/dotnet/msbuild/blob/main/src/Tasks/ManifestUtil/Util.cs)):

```csharp
if (string.IsNullOrEmpty(targetFrameWorkVersion) ||
    CompareFrameworkVersions(targetFrameWorkVersion, Constants.TargetFrameworkVersion40) <= 0)
{
    hashAlg = SHA1.Create();     // ← parameterless UpdateFileInfo() lands here
}
else
{
    hashAlg = SHA256.Create();
}
```

`UpdateFileInfo()` → `UpdateFileInfoImpl(null)` → `UpdateFileReference(f, null)` → `GetFileInfoImpl(path, null, …)`, so `string.IsNullOrEmpty(null)` is `true` and SHA-1 is selected.

I verified this against `Microsoft.Build.Tasks.Core` 17.14.8. Hashing a 15-byte file produced `rWbF5cgBBKfwb+yRL/iWJ9uBUaI=` 28 base64 chars, i.e. 20 bytes. `SHA1` of that content matches exactly; `SHA256` (`wb9/g5YunpZ/mH6E/IdlB/r3gjqCQ9SVAQZMFJ5ji9Q=`, 44 chars) does not.

kartheekp-ms: > UpdateFileInfo() hashes files at their ResolvedPath, which does not include .deploy; the suffixes must be absent when hashes are computed

If the payload files are signed in the temporary staging directory, will `ResolvedPath` point to the staging directory or to the original source directory?

Walking the steps as written:

- Line 236 calls `ApplicationManifest.ResolveFiles()`, resolving against the application manifest's directory — i.e. the **source** location.
- Line 237 copies those files to a **temporary** directory.
- Line 240 signs the payload files (in staging, per the copy-then-sign model).
- Line 241 calls `UpdateFileInfo()`, which hashes each reference at its `ResolvedPath`.

My reading of `ManifestUtil` is that `ResolvedPath` is a plain string property set once during `ResolveFiles()`, `ResolvePath` returns `Path.GetFullPath(...)` of the first search directory containing the file, and nothing re-points it afterward. If that's right, then at step 241 it still points at the source copy, and the recorded hashes would describe the **unsigned** originals rather than the signed staged files that get copied back in step 244.

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,290 @@
...
+
+### With `--no-sign-clickonce-deps`
+
+When `--no-sign-clickonce-deps` is specified, Sign CLI will update and sign only the explicitly provided manifest files without signing their dependencies:
+
+1. Before processing any file, obtain or wait for its coordinated signing operation.
+1. For each file provided by the user:
+   - If the file has a `.vsto` or `.application` file extension, read it as a deployment manifest, call `DeployManifest.ResolveFiles()` and `DeployManifest.UpdateFileInfo()` to update its metadata based on the current state of the referenced application manifest, then sign only the deployment manifest.
+   - If the file has a `.manifest` file extension, read it as an application manifest and call `ApplicationManifest.ResolveFiles()`. Temporarily remove `.deploy` suffixes from referenced payloads, call `ApplicationManifest.UpdateFileInfo()`, restore the suffixes, then sign only the application manifest.
+   - For other file types, apply the standard signing logic.
```
kartheekp-ms, line 281: If both the application manifest and the deployment manifest are passed in a single invocation, in what order are they processed? My understanding is that the application manifest must be signed before the deployment manifest is updated and signed, because the deployment manifest records the application manifest's hash, size, and assembly identity, all of which change when it is signed.

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,290 @@
...
+1. Sign payload files.
+1. Call `ApplicationManifest.UpdateFileInfo()` to refresh payload hashes, sizes, and identities, then restore the `.deploy` suffixes. (`UpdateFileInfo()` hashes files at their `ResolvedPath`, which does not include `.deploy`; the suffixes must be absent when hashes are computed.) Make the signed payload results available to waiting callers and complete their coordinated signing operations. Then sign the application manifest, make its signed result available, and complete its operation.
+1. Signing the application manifest changes its signed bytes and can change its assembly identity, including its public key token. Call `DeployManifest.ResolveFiles()` to re-resolve file references, then call `DeployManifest.UpdateFileInfo()` to refresh the deployment manifest's hash, size, and entry-point identity before signing the deployment manifest. When signing the deployment manifest with `mage.exe -update`, the `-appm` parameter updates the entry point reference to the application manifest. Make the signed deployment manifest available and complete its operation.
```
kartheekp-ms: > Signing the application manifest changes its signed bytes and can change its assembly identity, including its public key token. Call DeployManifest.ResolveFiles() to re-resolve file references, then call DeployManifest.UpdateFileInfo() to refresh the deployment manifest's hash, size, and entry-point identity

[`Manifest.UpdateFileInfo()`](https://github.com/dotnet/msbuild/blob/e45cc3de8a44d5b92cdad6e0ca7f8a5852c2afbd/src/Tasks/ManifestUtil/Manifest.cs#L581-L585) does not refresh entry-point identity. It only populates identity when it's currently `null`:

```csharp
// Manifest.UpdateAssemblyReference
if (a.AssemblyIdentity == null)
{
    a.AssemblyIdentity = AssemblyIdentity.FromFile(a.ResolvedPath);   // ← gated
}
UpdateFileReference(a, targetFrameworkVersion);                        // hash + size only
```

Since a published deployment manifest already carries an `assemblyIdentity` for its entry point, this block is skipped and the stale `publicKeyToken` is retained. Is my understanding, correct?

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Good catch, @kartheekp-ms! The parameterless `UpdateFileInfo()` overload computes SHA-1 hashes.

I verified that Sign CLI’s packaged `mage.exe` maps `-a sha256RSA` to `UpdateFileInfo("v4.5")`, and experimentally confirmed that it emits SHA-256 hashes. I updated the spec so version 2 consistently passes `"v4.5"`, preserving Sign CLI’s current SHA-256 behavior. The spec also clarifies that `"v4.5"` selects MSBuild’s hashing behavior rather than representing the application’s target framework.

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,292 @@
...
+1. Ensure [`Manifest.ReadOnly`](https://learn.microsoft.com/dotnet/api/microsoft.build.tasks.deployment.manifestutilities.manifest.readonly?view=msbuild-17-netcore) is `false` so the deployment manifest can be updated.
+1. Call [`DeployManifest.ResolveFiles()`](https://learn.microsoft.com/dotnet/api/microsoft.build.tasks.deployment.manifestutilities.manifest.resolvefiles?view=msbuild-17-netcore#microsoft-build-tasks-deployment-manifestutilities-manifest-resolvefiles) to resolve file references relative to the deployment manifest's directory. Preserve the resolved relative paths (including `.deploy` suffixes) when staging files.
```
kartheekp-ms: > Preserve the resolved relative paths (including `.deploy` suffixes) when staging files

Beginner question about "preserve the resolved relative paths (including `.deploy` suffixes)".

If Visual Studio published the payload files with the `.deploy` extension, the manifest references `App.dll` but the file on disk is `App.dll.deploy`. In that case, what does `ResolveFiles()` return for that reference?

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Thanks, @kartheekp-ms. You’re right that the steps as previously written did not ensure that `UpdateFileInfo(...)` would use the staged files rather than the source files.

The current implementation avoids this by re-reading the manifests from staging and resolving the new manifest objects against the staged directories, but the spec did not make the required outcome explicit. I updated it to require every `ResolvedPath` used by `UpdateFileInfo(...)` to identify the corresponding staged file, including the signed staged application manifest referenced by the deployment manifest. How those staged paths are established remains an implementation detail.

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Thanks, @kartheekp-ms. When `--no-sign-clickonce-deps` is used and both the application manifest and its deployment manifest are explicitly provided in the same invocation, the application manifest must be updated and signed first, regardless of input order or parallel scheduling.

I updated the spec to make this ordering explicit.

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Good catch. `DeployManifest.UpdateFileInfo(...)` refreshes the entry-point hash and size, but not an existing identity. I've updated the spec to require identity synchronization separately.

docs/specs/ClickOnce-Signing-Algorithm.md
dtivel: Good question. `ResolveFiles()` does not account for ClickOnce file-extension mapping, so a manifest reference such as `App.dll` will not resolve when the published file is `App.dll.deploy`. I've clarified that the mapped suffix must be preserved during staging, removed temporarily from the staged copy for resolution and hashing, then restored afterward.

docs/specs/ClickOnce-Signing-Algorithm.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Appendix A:  Signing algorithm version 1
+
+In a temporary directory:
+
+1. [[source](https://github.com/dotnet/sign/blob/d4a580a9232e9d7aac931ea57b844e87e255af9a/src/Sign.Core/Signer.cs#L137-L149)]  Copy the deployment manifest to a random file name with the same file extension (`.application` or `.vsto`).
+1. [[source](https://github.com/dotnet/sign/blob/d4a580a9232e9d7aac931ea57b844e87e255af9a/src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs#L261-L274)]  Copy all files from the deployment manifest's source directory and all its subdirectories to the temporary directory, while preserving the source's directory structure. _Because copying does not filter down to manifests and payload files, this step can result in overcopying._
+1. [[source](https://github.com/dotnet/sign/blob/d4a580a9232e9d7aac931ea57b844e87e255af9a/src/Sign.Core/DataFormatSigners/ClickOnceSigner.cs#L97-L113)]  Sign all `.deploy` and `.exe` files included by user's file matching patterns. _Previous overcopying can lead to oversigning in this step._
```
kartheekp-ms, line 172: I'm curious whether there's a way for customers to exclude certain files from signing in v2.

Line 245 says only, "Sign payload files", with no mention of the user's file matching patterns. Appendix A lines 170-171 describe v1 filtering payloads through those patterns — and line 171 explicitly handles the files they *exclude*. Line 69 also says "Implicit ClickOnce dependency traversal and user file matching remain separate", which I read as meaning the traversal signs every referenced payload regardless of the patterns.

Is that understanding, correct? If so, isn't it a breaking change for anyone who currently excludes third-party or already-signed assemblies?

dtivel: Good catch. Version 2 should preserve the existing ability to exclude third-party or already-signed payloads. I updated the spec so every referenced payload is still discovered, staged, and included when refreshing manifest metadata, but only supported payloads selected by the user's file-matching patterns are signed. Excluded payloads remain unchanged. The same matching behavior applies to referenced `Launcher.exe` files and adjacent executables.

kartheekp-ms (approved):

Issue #824, opened 2025-03-03
VSIX signing failed which contains a file with extension ".app"


StGabriel:
**Describe the bug**
VSIX signing does not work if there is an ".app" file inside.

**Repro steps**
I have a `Application.app` file inside the VSIX extension and the content is a JSON file:
```
{
  "_Name": "TestApplication"
}
```

**Expected behavior**
Signing will not fail. The deprecated and legacy version of [Microsoft's vsixsigntool](https://www.nuget.org/packages/Microsoft.VSSDK.Vsixsigntool) can sign the VSIX package.

**Actual behavior**
The VSIX signing fails with the output
```
[...]
info: Sign.Core.IDataFormatSigner[0]
      Signing SignTool job with 1 files.
trce: Sign.SignatureProviders.CertificateStore.CertificateStoreService[0]
      Fetching certificate from certificate store.
trce: Sign.SignatureProviders.CertificateStore.CertificateStoreService[0]
      Fetched certificate from certificate store. [0.4985 ms]
info: Sign.Core.IDataFormatSigner[0]
      Signing C:\Temp\7\whlkhfbq.w00\Application.app.
trce: Sign.Core.IDataFormatSigner[0]
      Getting SIP Data
trce: Sign.Core.IDataFormatSigner[0]
      Calling SignerSignEx3
fail: Sign.Core.IDataFormatSigner[0]
      Signing failed with error -2146762749.
info: Sign.Core.IDataFormatSigner[0]
      Performing attempt #2 of 3 attempts after 5 s.
info: Sign.Core.IDataFormatSigner[0]
      Signing C:\Temp\7\whlkhfbq.w00\Application.app.
trce: Sign.Core.IDataFormatSigner[0]
      Getting SIP Data
trce: Sign.Core.IDataFormatSigner[0]
      Calling SignerSignEx3
fail: Sign.Core.IDataFormatSigner[0]
      Signing failed with error -2146762749.
info: Sign.Core.IDataFormatSigner[0]
      Performing attempt #3 of 3 attempts after 11.1803398 s.
info: Sign.Core.IDataFormatSigner[0]
      Signing C:\Temp\7\whlkhfbq.w00\Application.app.
trce: Sign.Core.IDataFormatSigner[0]
      Getting SIP Data
trce: Sign.Core.IDataFormatSigner[0]
      Calling SignerSignEx3
fail: Sign.Core.IDataFormatSigner[0]
      Signing failed with error -2146762749.
[...]

PR #652, opened 2024-01-30
Support signing Dynamics Business Central App files

aholstrup1:
Support signing Dynamics Business Central App (.app) files

Support for signing Dynamics Business Central files is done by: 
* Downloading the SIP for signing .app files (NavSip.dll)
* Updating the wintrust.dll.ini file to include an entry for NavSip.dll
* Adding .app files as a supported file extension

Fixes #173

PR #846, opened 2025-04-03
Improve detection of Dynamics Business Central `.app` files

dtivel:
Resolve https://github.com/dotnet/sign/issues/824.

https://github.com/dotnet/sign/pull/652 added support for signing Dynamics Business Central `.app` files, but the `.app` file extension is not exclusive to Dynamics Business Central apps.  In this bug, Sign CLI fails to sign a JSON file with an `.app` file extension.

This change updates Sign CLI to be smarter about detecting Dynamics Business Central `.app` files.  These files begin with a 4-byte header `4e 41 56 58` (NAVX), so we should only try signing if an `.app` file starts with this header.

CC @aholstrup1

### DIFF
--- a/src/Sign.Core/DataFormatSigners/AzureSignToolSigner.cs
+++ b/src/Sign.Core/DataFormatSigners/AzureSignToolSigner.cs
@@ -15,7 +15,7 @@ namespace Sign.Core
         private readonly ICertificateProvider _certificateProvider;
         private readonly ISignatureAlgorithmProvider _signatureAlgorithmProvider;
         private readonly ILogger<IDataFormatSigner> _logger;
-        private readonly HashSet<string> _supportedFileExtensions;
+        private readonly IReadOnlyList<ISignableFileType> _signableFileTypes;
         private readonly IToolConfigurationProvider _toolConfigurationProvider;
 
         // Dependency injection requires a public constructor.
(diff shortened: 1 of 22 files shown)


src/Sign.Core/DataFormatSigners/AzureSignToolSigner.cs
```diff
@@ -36,45 +36,55 @@ public AzureSignToolSigner(
...
-                ".psd1",        // PowerShell data files
-                ".psm1",        // PowerShell module files
-                ".stl",
-                ".sys",
-                ".vbs",
-                ".vxd",
-                ".winmd"
+                // For PowerShell file extensions, see https://github.com/PowerShell/PowerShell/blob/2f4f585e7fe075f5c1669397ae738c554fa18391/src/System.Management.Automation/security/SecurityManager.cs#L97C1-L106C10
+                new SignableFileTypeByExtension(
+                    ".appx",
```
dtivel, line 43: `.app` removed here and added inside the `DynamicsBusinessCentralAppFileType` class.

joelverhagen:
Awesome change. Seems like we can closely move extensions to the dynamic file type as we run into similar bugs. I think the magic byte detection will go a long way.

Is there an escape hatch for the user, before a bug fix like this, so say "myjson.app" should not be signed?

joelverhagen (approved):

src/Sign.Core/DataFormatSigners/DynamicsBusinessCentralAppFileType.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    internal sealed class DynamicsBusinessCentralAppFileType : ISignableFileType
+    {
+        private const string FileExtension = ".app";
+
+        private readonly ReadOnlyMemory<byte> _expectedHeader;
+
+        internal DynamicsBusinessCentralAppFileType()
+        {
+            _expectedHeader = new byte[] { 0x4e, 0x41, 0x56, 0x58 }; // NAVX
```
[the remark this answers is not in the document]
dtivel: I don't think so.  I figured this out on my own then confirmed with people on the Dynamics Business Central team.

dtivel:
Thanks.  Yes, there was/is an escape hatch:  using glob patterns.  However, [that approach is not well documented](https://github.com/dotnet/sign/issues/783), so I don't see users figuring it out quickly/easily.

joelverhagen:
> Thanks. Yes, there was/is an escape hatch: using glob patterns. However, [that approach is not well documented](https://github.com/dotnet/sign/issues/783), so I don't see users figuring it out quickly/easily.

It might be worth giving a minimal, thrown together in that issue description prior to full documentation. If it only takes a minute. Just a thought!

thloke (approved):

aholstrup1 (approved):

src/Sign.Core/DataFormatSigners/AzureSignToolSigner.cs
```diff
@@ -36,45 +36,55 @@ public AzureSignToolSigner(
...
+                    ".appxbundle",
+                    ".cab",
+                    ".cat",
+                    ".cdxml",       // PowerShell cmdlet definition XML
+                    ".dll",
+                    ".eappx",
+                    ".eappxbundle",
+                    ".emsix",
+                    ".emsixbundle",
+                    ".exe",
```
clairernovotny, line 53: With this new model, why leave PE files as detected by extensions? Seems like a PeFileType can detect the PE headers similarly?

clairernovotny (changes requested):

src/Sign.Core/DataFormatSigners/DynamicsBusinessCentralAppFileType.cs
```diff
@@ -0,0 +1,40 @@
...
+            using (FileStream stream = file.OpenRead())
+            {
+                var header = new byte[_expectedHeader.Length];
+
+                if (stream.Read(header, offset: 0, header.Length) != header.Length)
+                {
+                    return false;
+                }
+
+                return header.AsSpan().SequenceEqual(_expectedHeader.Span);
```
jeffkl, line 36: Just for my own understanding, is there a need to use Span here?  Both types are already fixed length arrays with 4 elements.

src/Sign.Core/DataFormatSigners/DynamicsBusinessCentralAppFileType.cs
```diff
@@ -0,0 +1,40 @@
...
+        private const string FileExtension = ".app";
+
+        private readonly ReadOnlyMemory<byte> _expectedHeader;
+
+        internal DynamicsBusinessCentralAppFileType()
+        {
+            _expectedHeader = new byte[] { 0x4e, 0x41, 0x56, 0x58 }; // NAVX
+        }
+
+        public bool IsMatch(FileInfo file)
```
jeffkl, line 18: How would any I/O or file access error manifest in the tool?  Not sure if this method should handle those and bubble up a friendly message, if the caller would handle that, or if the tool would just crash.

jeffkl (approved):

src/Sign.Core/DataFormatSigners/AzureSignToolSigner.cs
dtivel: One step at a time.  That can be another PR.

src/Sign.Core/DataFormatSigners/DynamicsBusinessCentralAppFileType.cs
dtivel: This PR does not change how I/O errors are reported.  Both before and after this change, the tool would fail with a message like "Could not find file '___'." and a callstack and a failing exit code.

There is room for fine tuning here, either dealing with files which are inaccessible (say, due to permissions) or in how we report the errors, but we haven't received user feedback in this area yet.

joelverhagen (approved):

jeffkl (approved):
:shipit:

PR #862, opened 2025-05-08
Suppress CodeQL.SM05137

dtivel:
Resolve https://github.com/dotnet/sign/issues/861.

### DIFF
--- a/src/Sign.Cli/AzureCredentialOptions.cs
+++ b/src/Sign.Cli/AzureCredentialOptions.cs
@@ -71,6 +71,8 @@ namespace Sign.Cli
             }
 
             DefaultAzureCredentialOptions options = CreateDefaultAzureCredentialOptions(context.ParseResult);
+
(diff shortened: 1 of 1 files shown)


erdembayar (approved):

dtivel:
It's a warning generated by internal code quality scanners that basically recommends against using `DefaultAzureCredential` in services and instead preferring specific authentication strategies (like `ManagedIdentityCredential` and `WorkloadIdentityCredential`) to ensure highly deterministic behavior.

It doesn't apply here because Sign CLI isn't a service.

Still, it might make sense at some point to introduce distinct paths for `ManagedIdentityCredential` and `WorkloadIdentityCredential` for users who want to opt in to exactly that strategy.

joelverhagen (approved):

dtivel:
@dlemstra, I do think that's a step in the right direction.  If you want to use strategy X, we should use only strategy X.  `DefaultAzureCredential` may still be a fallback.

PR #864, opened 2025-05-10
Fix bug where the RSA object that is created with GetRSAPublicKey is not disposed.

### LINKED ISSUE #863: RSATrustedSigning is not disposing the RSA object created with GetRSAPublicKey
The RSATrustedSigning class is not disposing the RSA object that is created with GetRSAPublicKey in the PublicKey property.

dlemstra:
Fixes #863.

### DIFF
--- a/src/Sign.SignatureProviders.TrustedSigning/RSATrustedSigning.cs
+++ b/src/Sign.SignatureProviders.TrustedSigning/RSATrustedSigning.cs
@@ -15,7 +15,7 @@ namespace Sign.SignatureProviders.TrustedSigning
         private readonly CertificateProfileClient _client;
         private readonly string _accountName;
         private readonly string _certificateProfileName;
-        private readonly X509Certificate2 _certificate;
+        private readonly RSA _rsaPublicKey;
 
         public RSATrustedSigning(
             CertificateProfileClient client,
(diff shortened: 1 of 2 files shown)


src/Sign.SignatureProviders.TrustedSigning/RSATrustedSigning.cs
```diff
@@ -15,7 +15,7 @@ internal sealed class RSATrustedSigning : RSA
         private readonly CertificateProfileClient _client;
         private readonly string _accountName;
         private readonly string _certificateProfileName;
-        private readonly X509Certificate2 _certificate;
+        private readonly RSA _rsaPublicKey;
```
erdembayar, line 18: X509Certificate2 is not specifically only for RSA, does it break anything now?

dlemstra: We were doing `_certificate.GetRSAPublicKey()!` in the property I moved that to the constructor to make sure it's disposed after we are done using it. Functionally nothing was changed.

erdembayar (approved):

dtivel (approved):

dtivel:
Thanks, @dlemstra!

PR #866, opened 2025-05-10
Added RSAKeyVaultWrapper that resolves calls to the ExportParameters method of the RSAKeyVault.

### LINKED ISSUE #865: Clickonce singing is broken when using the KeyVaultService that creates a RSAKeyVault
Signing a Clickonce application with a key vault does not work because `RSAKeyVault` does not have an implementation for `ExportParameters` that retrieves the certificate from the key vault.  We could remove calls to `ExportParameters` in our code but that would not resolve methods where we pass in the `RSA` object and that code calls that method. One example where that is happening is inside the `SignedXml` class. To solve this we would need to return an `RSA` class that has a similar way of solving this problem like we do inside the `RSATrustedSigning` class.

dlemstra:
Fixes #865.

### DIFF
--- a/src/Sign.SignatureProviders.KeyVault/KeyVaultService.cs
+++ b/src/Sign.SignatureProviders.KeyVault/KeyVaultService.cs
@@ -85,7 +85,8 @@ namespace Sign.SignatureProviders.KeyVault
         {
             using X509Certificate2 certificate = await GetCertificateAsync(cancellationToken);
             RSAKeyVault rsaKeyVault = await _cryptographyClient.CreateRSAAsync(cancellationToken);
-            return new RSAKeyVaultWrapper(rsaKeyVault, certificate);
+            RSA rsaPublicKey = certificate.GetRSAPrivateKey()!;
+            return new RSAKeyVaultWrapper(rsaKeyVault, rsaPublicKey);
         }
     }
 }
--- a/src/Sign.SignatureProviders.KeyVault/KeyVaultService.cs
+++ b/src/Sign.SignatureProviders.KeyVault/KeyVaultService.cs
@@ -83,7 +83,9 @@ namespace Sign.SignatureProviders.KeyVault
 
         public async Task<RSA> GetRsaAsync(CancellationToken cancellationToken)
         {
-            return await _cryptographyClient.CreateRSAAsync(cancellationToken);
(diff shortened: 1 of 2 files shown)


src/Sign.SignatureProviders.KeyVault/KeyVaultService.cs
```diff
@@ -83,7 +83,9 @@ public async Task<X509Certificate2> GetCertificateAsync(CancellationToken cancel
 
         public async Task<RSA> GetRsaAsync(CancellationToken cancellationToken)
         {
-            return await _cryptographyClient.CreateRSAAsync(cancellationToken);
+            using X509Certificate2 certificate = await GetCertificateAsync(cancellationToken);
+            RSAKeyVault rsaKeyVault = await _cryptographyClient.CreateRSAAsync(cancellationToken);
+            return new RSAKeyVaultWrapper(rsaKeyVault, certificate);
```
dtivel, line 88: I suspect that a lack of clarity around ownership led to a bug like [this](https://github.com/dotnet/sign/pull/864/files#diff-7d0f486dde88e84ae154d391a1e664a07e0834c7b80f66b2a7ef586301384b80R94-R95).

A [coding guideline around `IDisposable` ownership](https://github.com/dotnet/sign/issues/869) would likely drive bugs like it down.  For example, [there is an analyzer rule for transferring ownership at a constructor.](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-quality-rule-options#dispose_ownership_transfer_at_constructor)

Here `rsaKeyVault` ownership is being transferred to `RSAKeyVaultWrapper`, while `certificate` ownership is not.  `RSAKeyVaultWrapper` only needs the certificate's `RSA`, so we could achieve local consistency by moving the `GetRSAPublicKey()` call out of `RSAKeyVaultWrapper` and into this method and passing the `RSA` instance instead.

Still, I'm sure there's a lack of global consistency throughout the repo, so I wouldn't hold up this PR on this.  For example, the "transfer ownership at constructor" rule suggested here isn't applied [here](https://github.com/dotnet/sign/pull/864/files#diff-7d0f486dde88e84ae154d391a1e664a07e0834c7b80f66b2a7ef586301384b80R94-R95).

dtivel (approved):

dtivel (approved):

joelverhagen (approved):

Issue #775, opened 2024-10-14
Option to not sign container contents


zooba:
**Is your feature request related to a problem? Please describe.**
I have a complex build where I produce one set of binaries and multiple containers. I want all the binaries to be signed once so they are identical in each container. This tool always signs the contents of my Nuget and MSIX containers, even though they are already signed.

**Describe the solution you'd like**
A command line option to only sign the specified container itself, and not extract it and sign every file inside of it.

Updating the appxmanifest is nice, though, but I'd happily provide a second option to do that. My build already handles updating it manually.

**Describe alternatives you've considered**
An environment variable would also be fine.

**Additional context**
Include the output of sign --version.

```
0.9.1-beta.24469.1+5516908609085978174904945d54144c117041dc
```
(On my local machine, not my CI machine, but it's installed fresh every time there. I'm not running an official build to get a version number for a feature request - I've already checked that there's no code in `main` that would do what I need.)

dlemstra:
This sounds like a reasonable addition but this should probably be an option instead of a check for an environment variable.

Issue #647, opened 2023-11-03
Unable to skip already signed files (for instance for publishing a website zip)


CZEMacLeod:
It would be useful when signing e.g. a zip file created by publishing a website, to only sign unsigned files. This way existing signatures on DLLs such as those distributed by Microsoft don't get overwritten, and there would be less overhead when using KeyVault HSM.

This might be indirectly related to #547 where the default action was changed to overwrite, and the fact that the documented `force` flag doesn't seem to exist. The options to skip, overwrite, or append a signature would seem to be appropriate instead of force.

If you select to skip existing signatures, it should (optionally?) check the existing signature is valid, then skip the file if it is. If it is not, then it should 'fall through' to overwrite. In the case where append is used, it should also check the existing signature.

At the moment, the only way of implementing this that I can see, is to pre-scan the files and generate a filelist, although it is not entirely clear how this would work with files inside nuget or zip files as the documentation around the --file-list command is a little unclear.

Version: 0.9.1-beta.23530.1+f4efed9e8fb3296f29497b90feb6548e506f2078

dtivel:
CC @clairernovotny

Authenticode supports multiple signatures.  It was planned that Sign CLI's default behavior was to append new signatures. 
 However, last we checked a Windows API we use (`SignerSignEx3(...)`) did not support appending; instead, it overwrote an existing signature.  The `--force` option in the draft specification is in a limbo state because we haven't a good solution yet.

That said, there is a way to skip already signed files using file lists.  You would use a file list to exclude already signed files in a separate invocation of Sign CLI.  File lists have one file pattern per line.  Each pattern can be a glob pattern.

CZEMacLeod:
@dtivel Yes - I had worked out that the filelist might be a way of doing it, however, I run the command on a directory with multiple .nuget files (sometimes up to 100), and to work out the list would have to open and scan each file in the archive to check if it is already signed and add the ones that aren't to the list. 
I'm not even sure how it would work with the multiple nuget files. Would I need to include a pattern to sign the nuget package itself too?
If this is the only available mechanism, then perhaps a good example of how to do this would be an idea?

As the command is already opening and unzipping the files to sign them, it would be nice to be able to test the files to check if they already have a valid signature, and only sign those without the signature. I appreciate that append and/or overwrite were the original intention - however, adding the third option to skip would be useful.

I appreciate this would be an extra step in the logic, and might slow it down somewhat. However, it would be much easier than basically doing the same work twice.

[...]

PR #873, opened 2025-05-13
Treat MSI and CAB files as containers

### LINKED ISSUE #673: Sign all unsigned assemblies in a .msi
**Is your feature request related to a problem? Please describe.**
Currently, when signing a .msi, it will correctly sign that msi file but none of the assemblies inside are signed. When I attempt to use --file-list to manually specify the names of the assemblies to sign, I don't see them getting a digital signature. I'm using a WiX project to create the .msi file. This happens during dotnet build. I don't want signing to have to get configured within every project and solution by all the developers in my company. They should not have to focus on that framework that we can automate.

**Describe the solution you'd like**
The easiest solution would be an option --recursive-signing that would check the content of the .msi and sign every file (or dll) that doesn't already have a signature. I'm uncertain if unzipping, looping, signing, and rezipping things back into a .msi won't cause corrupted installers though. That would likely need to be considered. Unless the internal Win32 Signing Invoke call used to sign the .msi has a flag or option to already handle such recursive signing?

**Describe alternatives you've considered**
If file-list would have worked, I would have made a pipeline stage that after dotnet build, checks every project to find their created .dll and then add the names to a new text file to use during dotnet sign.

Current workarounds seem to be faffing around with post and pre-build actions (which I'd love to avoid). I've read one blog post of someone attempting to build every project except the WiX first, sign the assemblies then build the WiX but they indicated that WiX rebuilds everything and overwrites the signed assemblies with unsigned ones again.

**Additional context**
sign --version 0.9.1-beta.24170.3+394fc50bebb3ddaa3f79fc975f552d4ea2787552

### LINKED ISSUE #874: Treat CAB files as containers
**Is your feature request related to a problem? Please describe.**
The contents of CAB files aren't signed.

**Describe the solution you'd like**
The contents of CAB files should be treated the same as in ZIP files. They should be recursively extracted, signed, and re-assembled before the main CAB file is signed.

**Describe alternatives you've considered**
None

**Additional context**
None

bricelam:
Fixes #673, fixes #874

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -6,6 +6,8 @@
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <!-- TODO: Mirror WixToolset.Dtf.WindowsInstaller.Package and dependencies -->
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
   <disabledPackageSources />
 </configuration>
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -5,6 +5,8 @@
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
+    <!-- TODO: Mirror WixToolset.Dtf.* and dependencies and remove -->
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
   <packageSourceMapping>
     <clear />
@@ -33,6 +35,10 @@
     </packageSource>
     <packageSource key="dotnet-tools">
       <package pattern="sn" />
+    </packageSource>
+    <packageSource key="nuget.org">
+      <!-- TODO: Mirror WixToolset.Dtf.* and dependencies and move -->
+      <package pattern="WixToolset.*" />
     </packageSource>
   </packageSourceMapping>
   <disabledPackageSources>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -33,5 +33,6 @@
     <PackageVersion Include="System.Security.Cryptography.Pkcs" Version="9.0.4" />
     <PackageVersion Include="System.Security.Cryptography.Xml" Version="9.0.4" />
     <PackageVersion Include="System.Text.Json" Version="9.0.4" />
+    <PackageVersion Include="WixToolset.Dtf.WindowsInstaller.Package" Version="6.0.0" />
   </ItemGroup>
 </Project>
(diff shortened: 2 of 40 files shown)


NuGet.Config
```diff
@@ -6,6 +6,8 @@
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <!-- TODO: Mirror WixToolset.Dtf.WindowsInstaller.Package and dependencies -->
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
dlemstra, line 10: This should be resolved before this PR is merged.

bricelam, line 10: Do this (and revert this file change) before merging

bricelam:
I wasn't sure what to do about testing here. I've tested it manually, of course. But generating an MSI in the tests felt like overkill and would require significant new dependencies. I also considered checking in a test MSI, but wasn't sure you guys would go for that. Let me know if (and how) you want me to add more tests.

NuGet.Config
bricelam: Here are the instructions for [mirroring a package](https://github.com/dotnet/arcade/blob/main/Documentation/MirroringPackages.md). Also be sure to ~~revert~~ **update** this file.

bricelam:
Fixing this will probably make #647 ~~and #775~~ (resolved) more urgent for some users.

Foda:
> @bricelam I was able to verify this manually, but this only works if the `.cab` your `.msi` uses is embedded. It causes a regression with `.msi` installers with non-embedded `.cab` files (ie: #875). It throws in the wixtoolset code and appears as a "file not found" error in a temp folder.

Just a follow-up: it's probably OK for this regression as long as it doesn't roll out in a release. We'd like to add a new flag that doesn't recurse into file containers to go along with this.

kartheekp-ms:
@bricelam please rebase since we migrated tests from [Moq to NSubstitute](https://github.com/dotnet/sign/pull/1054).

bricelam:
PR updated. This should be good to go now. MSI tests added.

bricelam:
*From **Claude Code***

A note on test coverage, since one behavior here is less covered than the rest.

Two files are now signed as a unit: an MSI and any cabinet it owns. `Signer` fans out over its inputs in parallel, each in its own temporary directory, so before this change a glob matching both `foo.msi` and its `data1.cab` signed the cabinet twice -- once as part of the package, once as an input of its own -- racing on the output path. Because each pass produces its own timestamp countersignature, the two signed payloads differ, and the `File` table could end up describing a payload that was never shipped.

**What is covered deterministically:** `SigningDependencyProviderTests` covers the ownership resolution itself -- `ExcludeOwnedFiles` removing an owned file, handling an owner that is itself owned, and leaving unrelated files alone. `MsiContainerTests` and `MsiSigningDependencyReaderTests` cover extraction, re-packaging, the `File` table update, and the external-cabinet round trip against real packages built by `MsiCreator`. Removing the dependency copy fails 3 of the 4 end-to-end MSI tests, so the original external-cabinet regression is genuinely guarded.

**What is not:** `SignAsync_WhenFilesAreMsiAndItsCabinet_LeavesPackageDescribingItsPayload` asserts the right invariant -- the shipped payload matches the recorded `FileSize` and `MsiFileHash` -- but it still passes with the ownership pre-pass disabled. The package's own run does much more work (copy in, extract, sign, `UpdateFiles`, copy out) and so reliably wins the write race. The assertion is correct; it just can't be provoked by timing. Treat it as documentation of intent rather than a regression guard.

The deterministic version would assert the *decision* rather than the outcome: capture the container logger in the test service provider and assert that no container is ever opened for an owned cabinet. That flips reliably when the pre-pass is disabled, because it depends only on `ExcludeOwnedFiles`. I have not added it -- it needs a small recording `ILoggerProvider` in `Sign.TestInfrastructure` and a change to how `SignerTests` builds its service collection, and that felt worth agreeing on first.

Happy to add it if you would like it in this PR.

Issue #887, opened 2025-05-26
Logging:  noisy Azure logging


dtivel:
Since https://github.com/dotnet/sign/commit/1cd36786f5742ec6cdcee9d9e7406198b0f23788, the default logging level displays warnings and errors from Azure libraries.  These warnings and errors are internal and do not impact the overall functioning of Sign CLI.  While they may be useful at times, by default they are usually not useful and can be misleading.

Example https://github.com/dotnet/sign/issues/836

PR #808, opened 2024-12-11
Add logging through to Azure services

clairernovotny:
Refactored to use Azure services via DI as that also configures it to work with ILogger.

PR #888, opened 2025-05-26
Suppress noisy Azure logging

dtivel:
Resolve https://github.com/dotnet/sign/issues/887.

This change suppresses noisy internal warnings and errors from Azure libraries which were introduced by https://github.com/dotnet/sign/pull/808.

### DIFF
--- a/src/Sign.Cli/appsettings.json
+++ b/src/Sign.Cli/appsettings.json
@@ -1,8 +1,8 @@
 {
     "Logging": {
         "LogLevel": {
-            "Azure.Core": "Critical",
-            "Azure.Identity": "Critical"
+            "Azure.Core": "Error",
+            "Azure.Identity": "Error"
         }
     }
 }


erdembayar (dismissed):

src/Sign.Cli/appsettings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "Logging": {
+        "LogLevel": {
+            "Azure.Core": "Critical",
```
dlemstra, line 4: Should this not be `Error` instead?

dtivel: @dlemstra, changed to `Error`.

As far as other log levels go, can you create a separate issue and include specific suggestions, if you have any?

dlemstra: I will do that

jeffkl (approved):
I'm assuming this was manually verified?  Its a good idea to have before/after screenshots if possible that shows how output is affected by changes like this.

PR #891, opened 2025-05-26
Added --azure-credential-type option that can be used to specify the credential type for Azure authentication.

### LINKED ISSUE #867: Add option to specify the Azure Credential Type
It would be nice if a user could specify which credential type would be used instead of `DefaultAzureCredential`. I think we should call this option `--azure-credential-type` and allow the following values:
- azure-cli (`AzureCliCredential`)
- environment (`EnvironmentCredential`)
- managed-identity (`ManagedIdentityCredential`)

And when this option is not specified we default back to `DefaultAzureCredential` as we do now. I have created a proof of concept locally but not yet a PR because I think it would probably be a good idea to discuss first which options we want to allow. The use case for this would be a GitHub or Azure DevOps pipeline so I don't think we would need to add support for something like `VisualStudioCredential`.

dlemstra:
Fixes #867.

### DIFF
--- a/src/Sign.Cli/AzureCredentialOptions.cs
+++ b/src/Sign.Cli/AzureCredentialOptions.cs
@@ -13,6 +13,11 @@ namespace Sign.Cli
 {
     internal sealed class AzureCredentialOptions
     {
+        internal Option<string?> CredentialTypeOption = new Option<string?>(["--azure-credential-type", "-act"], Resources.CredentialTypeOptionDescription).FromAmong(
+            AzureCredentialType.AzureCli,
+            AzureCredentialType.AzurePowerShell,
+            AzureCredentialType.ManagedIdentity,
+            AzureCredentialType.WorkloadIdentity);
         internal Option<string?> ManagedIdentityClientIdOption = new(["--managed-identity-client-id", "-mici"], Resources.ManagedIdentityClientIdOptionDescription);
         internal Option<string?> ManagedIdentityResourceIdOption = new(["--managed-identity-resource-id", "-miri"], Resources.ManagedIdentityResourceIdOptionDescription);
         internal Option<bool?> ObsoleteManagedIdentityOption { get; } = new(["--azure-key-vault-managed-identity", "-kvm"], Resources.ManagedIdentityOptionDescription) { IsHidden = true };
@@ -22,6 +27,7 @@ namespace Sign.Cli
 
         internal void AddOptionsToCommand(Command command)
         {
+            command.AddOption(CredentialTypeOption);
             command.AddOption(ManagedIdentityClientIdOption);
             command.AddOption(ManagedIdentityResourceIdOption);
             command.AddOption(ObsoleteManagedIdentityOption);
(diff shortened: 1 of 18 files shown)


src/Sign.Cli/AzureCredentialType.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+namespace Sign.Cli
+{
+    internal static class AzureCredentialType
+    {
+        public const string AzureCli = "azure-cli";
+        public const string AzurePowerShell = "azure-power-shell";
```
jackmtpt, line 10:
```suggestion
        public const string AzurePowerShell = "azure-powershell";
```

dlemstra: I was also surprised it was done like that with `AzurePowerShellCredential` so that is why I decided to do it like that. Maybe @dtivel can decide what we should do here?

dtivel: `azure-powershell` is more consistent with my finding that when a product name is a compound word (like PowerShell, or MSBuild, DevOps), there's no hyphenation.

* MSBuild:  `dotnet msbuild` (not `ms-build`)
* Azure DevOps:  `az devops` (not `dev-ops`)

src/Sign.Cli/AzureCredentialOptions.cs
```diff
@@ -70,10 +76,38 @@ internal DefaultAzureCredentialOptions CreateDefaultAzureCredentialOptions(Parse
...
+                    {
+                        return new ManagedIdentityCredential(new ResourceIdentifier(managedIdentityResourceId));
+                    }
 
-            // CodeQL [SM05137] Sign CLI is not a production service.
-            return new DefaultAzureCredential(options);
+                    return new ManagedIdentityCredential();
+
+                case AzureCredentialType.WorkloadIdentity:
+                    return new WorkloadIdentityCredential();
```
dtivel, line 103: `WorkloadIdentityCredential` seems to require tenant ID, client ID, and a federated token file, but the only way to set those today is through environment variables.  This PR does not change that.  This PR is focused on discriminating credential types not adding options for them.  However, by elevating this and other credential types to an option, it highlights how inconsistent credential type handling is.  For example, `managed-identity` allows you to set client ID or resource ID from the command line, but:

* `azure-cli` doesn't allow you to set tenant ID and subscription.
* `azure-powershell` doesn't allow you to set tenant ID.
* `workload-identity` doesn't allow you to set tenant ID, client ID, and token file path.

In each of those cases, I just looked at the `-Options` parameter on the `-Credential` class constructor.

All this is to say that we might see requests to expose those arguments as additional options.

I've never tried some of these credential types myself.  Are there any you haven't tested?

dtivel (commented):
Overall, this looks great!  Thanks, @dlemstra.

src/Sign.Cli/AzureCredentialType.cs
dlemstra: I have updated the PR.

src/Sign.Cli/AzureCredentialOptions.cs
dlemstra: This PR is focused on doing the same as `DefaultAzureCredential` but only use the specified credential type. And right now everything is done through environment variables. I can understand we might need to add extra options but that should be something that we take a look at when we get those kind of requests and they cannot use the environment variables. I have tested the `azure-cli` option locally because I will also be going to use that in my pipeline.

Foda (approved):

PR #1016, opened 2026-05-06
Honor content type overrides

### LINKED ISSUE #1015: Content type overrides in [Content_Types].xml are not honored
**Describe the bug**
Currently, the Overrides in the [Content_Types].xml file in the package are not being honored during signing.

For example, if the package contains two .txt files, 'file1.txt' and 'file2.txt', and the [Content_Types].xml defines the default ContentType for .txt as 'text/plain' but has an Override for 'file2.txt' to be 'text/xml', it will still use 'text/plain' when signing for 'file2.txt' instead of 'text/xml'.

**Repro steps**
This is an example vsix that will not honor the overrides properly: https://github.com/kennethr127/sign/blob/db8f4bc9336b04d1c851ac43a36a4344989e1d2e/test/Sign.Core.Test/TestAssets/VSIXSamples/OpenVsixSignToolTest-Overrides.vsix

1. Sign the package (Ex: `dotnet sign.dll code azure-key-vault "C:\Users\kereuter\Downloads\OpenVsixSignToolTest-Overrides.vsix" -kvu <KeyVaultURL> -kvc <CertName> -o "C:\Users\kereuter\Downloads\OpenVsixSignToolTest-Overrides-Signed.vsix" -act azure-cli`)
2. Use PackageDigitalSignatureManager.VerifySignatures to check the signature (script example seen below)

```
$packagePath = "C:\Users\kereuter\Downloads\OpenVsixSignToolTest-Overrides-Signed.vsix"
[Reflection.Assembly]::LoadWithPartialName('WindowsBase') | Out-Null
$packageToSign = [System.IO.Packaging.Package]::Open($packagePath)
$signatureManager = New-Object System.IO.Packaging.PackageDigitalSignatureManager($packageToSign)
$signatureManager.VerifySignatures($true)
$packageToSign.Close()
```

**Expected behavior**
The call to `$signatureManager.VerifySignatures($true)` returns "Success"

**Actual behavior**
The call to `$signatureManager.VerifySignatures($true)` returns "InvalidSignature"

kennethr127:
Currently, the Overrides in the [Content_Types].xml file in the package are not being honored during signing. This update addresses that issue by checking for any overrides when retrieving the ContentType.

Fixes #1015


kennethr127:
@dotnet-policy-service agree company="Microsoft"

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +188,24 @@ public void ShouldReturnSignatureForSignedPackage()
...
+                var partToCheck = new Uri("/extension.vsixmanifest", UriKind.Relative);
+                var part = package.GetPart(partToCheck);
+                Assert.True(part != null && part.ContentType == "text/xml");
+            }
+            using (var package = OpcPackage.Open(path, OpcPackageFileMode.ReadWrite))
+            {
+                var partToCheck = new Uri("/extension.vsixmanifest", UriKind.Relative);
+                var part = package.GetPart(partToCheck);
+                Assert.True(part != null && part.ContentType == "text/xml");
+            }
```
dtivel: @kennethr127, is the second open/assertion necessary? Nothing modifies the package between assertions, and `ContentType` does not depend on read/write mode. Opening `SamplePackageWithOverrides` once in read-only mode with `OpcPackage.Open(...)` should provide the same coverage while removing the shadow copy and duplicate verification.

src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
```diff
@@ -60,7 +60,10 @@ public string ContentType
             {
                 var extension = Path.GetExtension(_path)?.TrimStart('.');
 
-                return Package.ContentTypes.FirstOrDefault(ct => string.Equals(ct.Extension, extension, StringComparison.OrdinalIgnoreCase))?.ContentType ?? OpcKnownMimeTypes.OctetString;
+                var defaultContentType = Package.ContentTypes.FirstOrDefault(ct => string.Equals(ct.Extension, extension, StringComparison.OrdinalIgnoreCase));
+                var overrideContentType = Package.ContentTypes.FirstOrDefault(ct => string.Equals(ct.PartName?.Trim('/'), _path, StringComparison.OrdinalIgnoreCase));
```
dtivel: @kennethr127, use `TrimStart('/')` instead of `Trim('/')`. A trailing slash is invalid for an OPC part name, but removing it can silently make a malformed override match a different valid part. For example:

- `/folder/file.txt/` would incorrectly match `folder/file.txt`.
- `/extensionless/` would incorrectly match the part `extensionless`.

Removing only the required leading slash avoids accepting malformed override names. Also filter `Mode` by `OpcContentTypeMode.Override` and `Default` to make the intended precedence explicit.

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +188,24 @@ public void ShouldReturnSignatureForSignedPackage()
...
+            using (var package = OpcPackage.Open(path, OpcPackageFileMode.ReadWrite))
+            {
+                var partToCheck = new Uri("/extension.vsixmanifest", UriKind.Relative);
+                var part = package.GetPart(partToCheck);
+                Assert.True(part != null && part.ContentType == "text/xml");
```
dtivel: @kennethr127, use separate assertions so failures identify whether the part was missing or had the wrong content type:

```csharp
Assert.NotNull(part);
Assert.Equal("text/xml", part.ContentType);
```

`Assert.True(part != null && part.ContentType == "text/xml")` loses that diagnostic information.

dtivel (changes requested):
@kennethr127, thank you for contributing this fix!

I’m requesting a few focused changes before merging:

- Avoid stripping trailing slashes when matching override part names, since that can make malformed names match valid parts.
- Make the `Override` and `Default` modes explicit when selecting content types.
- Simplify the test to open the source VSIX once in read-only mode.
- Use separate `NotNull` and `Equal` assertions for clearer failures.

With those changes, this should be ready to merge.

src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
dlemstra: Might also be good to add some tests for those cases.

kennethr127: I updated `Trim` to `TrimStart` and added another test to cover this scenario.

I also added the filter to check for `OpcContentTypeMode.Override` and `OpcContentTypeMode.Default`.

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
kennethr127: That's a good point; I separated those Asserts and added some additional ones to make this more robust.

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
kennethr127: Yeah, the `ShadowCopyPackage` call was not really needed. I was mimicking one of the other test methods and didn't remove it. I updated this to only use the `Open` call instead.

kennethr127:
@dtivel Thank you reviewing this change and leaving feedback! I just pushed an update that should address all your feedback. Please let me know if I missed anything.

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +189,44 @@ public void ShouldReturnSignatureForSignedPackage()
...
         }
 
+        [Fact]
+        public void ShouldHonorNewContentTypeIfOverrideIsListedInContentTypesFile()
+        {
+            using (var package = OpcPackage.Open(SamplePackageWithOverrides, OpcPackageFileMode.ReadWrite))
+            {
+                var partToCheck = new Uri("/extension.vsixmanifest", UriKind.Relative);
+                var part = package.GetPart(partToCheck);
+                Assert.True(part != null);
```
dtivel:
```suggestion
                Assert.NotNull(part);
```

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +189,44 @@ public void ShouldReturnSignatureForSignedPackage()
...
+                Assert.False(part.ContentType == "text/plain");
+                Assert.True(part.ContentType == "text/xml");
+
+                partToCheck = new Uri("/extensionless", UriKind.Relative);
+                part = package.GetPart(partToCheck);
+                Assert.True(part != null);
```
dtivel:
```suggestion
                Assert.NotNull(part);
```

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +189,44 @@ public void ShouldReturnSignatureForSignedPackage()
...
+        [Fact]
+        public void ShouldNotHonorNewContentTypeIfInvalidOverrideIsListedInContentTypesFile()
+        {
+            using (var package = OpcPackage.Open(SamplePackageWithInvalidOverrides, OpcPackageFileMode.ReadWrite))
+            {
+                var partToCheck = new Uri("/extension.vsixmanifest", UriKind.Relative);
+                var part = package.GetPart(partToCheck);
+                Assert.True(part != null);
+                Assert.False(part.ContentType == "text/xml");
+                Assert.True(part.ContentType == "text/plain");
```
dtivel:
```suggestion
                Assert.NotNull(part);
                Assert.Equal("text/plain", part.ContentType);
```

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +189,44 @@ public void ShouldReturnSignatureForSignedPackage()
...
+
+                partToCheck = new Uri("/extensionless", UriKind.Relative);
+                part = package.GetPart(partToCheck);
+                Assert.True(part != null);
+                Assert.False(part.ContentType == "text/plain");
+                Assert.True(part.ContentType == "application/octet-stream");
```
dtivel:
```suggestion
                Assert.NotNull(part);
                Assert.Equal("application/octet-stream", part.ContentType);
```

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +189,44 @@ public void ShouldReturnSignatureForSignedPackage()
...
+        [Fact]
+        public void ShouldHonorNewContentTypeIfOverrideIsListedInContentTypesFile()
+        {
+            using (var package = OpcPackage.Open(SamplePackageWithOverrides, OpcPackageFileMode.ReadWrite))
+            {
+                var partToCheck = new Uri("/extension.vsixmanifest", UriKind.Relative);
+                var part = package.GetPart(partToCheck);
+                Assert.True(part != null);
+                Assert.False(part.ContentType == "text/plain");
+                Assert.True(part.ContentType == "text/xml");
```
dtivel:
```suggestion
                Assert.NotNull(part);
                Assert.Equal("text/xml", part.ContentType);
```

I recommend using `Assert.NotNull(part)` rather than `Assert.True(part != null)`, and `Assert.Equal(expected, actual)` rather than `Assert.True`/`Assert.False` for value comparisons. These purpose-built assertions produce clearer failure messages. In particular, when an equality assertion fails, xUnit includes both the expected and actual values, making failures easier to diagnose.

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +189,44 @@ public void ShouldReturnSignatureForSignedPackage()
...
+
+                partToCheck = new Uri("/extensionless", UriKind.Relative);
+                part = package.GetPart(partToCheck);
+                Assert.True(part != null);
+                Assert.False(part.ContentType == "application/octet-stream");
+                Assert.True(part.ContentType == "text/plain");
```
dtivel:
```suggestion
                Assert.NotNull(part);
                Assert.Equal("text/plain", part.ContentType);
```

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +189,44 @@ public void ShouldReturnSignatureForSignedPackage()
             }
         }
 
+        [Fact]
+        public void ShouldHonorNewContentTypeIfOverrideIsListedInContentTypesFile()
+        {
+            using (var package = OpcPackage.Open(SamplePackageWithOverrides, OpcPackageFileMode.ReadWrite))
```
dtivel: Read-only is appropriate.

```suggestion
            using (var package = OpcPackage.Open(SamplePackageWithOverrides))
```

test/Sign.Core.Test/Tools/VSIXSignTool/OpcPackageTests.cs
```diff
@@ -187,6 +189,44 @@ public void ShouldReturnSignatureForSignedPackage()
...
+                Assert.True(part != null);
+                Assert.False(part.ContentType == "application/octet-stream");
+                Assert.True(part.ContentType == "text/plain");
+            }
+        }
+
+        [Fact]
+        public void ShouldNotHonorNewContentTypeIfInvalidOverrideIsListedInContentTypesFile()
+        {
+            using (var package = OpcPackage.Open(SamplePackageWithInvalidOverrides, OpcPackageFileMode.ReadWrite))
```
dtivel:
```suggestion
            using (var package = OpcPackage.Open(SamplePackageWithInvalidOverrides))
```

dtivel (commented):
Thanks for the updates, @kennethr127!

dtivel:
@kennethr127, thanks for the changes. Can you rebase on latest `main` branch?

dtivel (approved):
LGTM.  Thanks, @kennethr127!

Issue #1033, opened 2026-07-01
Feature request: Export the signing certificate


jozefizso:
Add an option that writes the public information of the certificate used for a signing operation to a `.cer` file.

**Is your feature request related to a problem? Please describe.**
We publish signed nuget packages to NuGet Gallery. It requires us to manually upload the certificate to the account so Gallery can verify package validity.

We are using Azure Artifact Signing which rotates signing certificates daily. Our release scripts and automation are extracting the certificate information from signed nuget package using the `Knapcode.CertificateExtractor` tool.

**Describe the solution you'd like**
A new optional argument `--certificate-output <path-to-certificate.cer>` which will save the certificate to the file when the argument was set.

**Describe alternatives you've considered**
We must install the `Knapcode.CertificateExtractor` dotnet tool and use it to extract certificate data.

**GitHub Actions example**

```yaml
- name: Sign artifacts
  shell: pwsh
  run: >
    ./sign code artifact-signing
    **/*.nupkg
    --certificate-output "${{ runner.temp }}\signing-certificate.cer"

- name: Publish certificate path
  id: signing-certificate
  shell: pwsh
  run: '"certificate_file=${{ runner.temp }}\signing-certificate.cer" >> $env:GITHUB_OUTPUT'
```

Later steps can use `steps.signing-certificate.outputs.certificate_file` to inspect, archive, or upload the certificate required by NuGet Gallery.

**Benefits**

- Automation always receives the certificate that actually signed the artifacts.
- Certificate rotation does not require scripts to predict or separately retrieve the active certificate.
- The output is safe to share with downstream steps because it contains only the public certificate.

kartheekp-ms:
Team Triage: Your scenario should be addressed once https://github.com/NuGet/NuGetGallery/issues/10027 is implemented. Here is the [spec](https://github.com/NuGet/Home/blob/dev/accepted/2024/improve-support-for-packages-signed-with-trusted-signing.md). 

Please upvote the linked issue in NuGetGallery repo.

jozefizso:
I am sorry but Microsoft absolutely ignores the NuGet Gallery development and is unable to deliver a feature which would work for ATS customers.

The solution for everybody is the sign tool to export the exact certificate used to sign.

PR #1035, opened 2026-07-05
Implement new `--certificate-output` argument to save the certificate information

jozefizso:
The certificate information is useful when signing code with Azure Artifacts Signing
which rotates certificates each day. Scripts and automation can use correct certificate
used to sign content.

This helps with NuGet Gallery where we must manually upload certificates.

### DIFF
--- a/src/Sign.Cli/TrustedSigningCommand.cs
+++ b/src/Sign.Cli/TrustedSigningCommand.cs
@@ -114,18 +114,13 @@ namespace Sign.Cli
                 cancellationToken.ThrowIfCancellationRequested();
 
                 string? certificateOutput = parseResult.GetValue(CertificateOutputOption);
-                FileInfo? certificateOutputFile = null;
-
-                if (!string.IsNullOrEmpty(certificateOutput))
-                {
-                    DirectoryInfo baseDirectory = parseResult.GetValue(codeCommand.BaseDirectoryOption)!;
-                    certificateOutputFile = new FileInfo(CodeCommand.ExpandFilePath(baseDirectory, certificateOutput));
-
-                    if (certificateOutputFile.Exists)
-                    {
-                        throw new IOException($"The certificate output file already exists: '{certificateOutputFile.FullName}'.");
-                    }
-                }
+                DirectoryInfo baseDirectory = parseResult.GetValue(codeCommand.BaseDirectoryOption)!;
+                string? signingOutput = parseResult.GetValue(codeCommand.OutputOption);
+                FileInfo? certificateOutputFile = CodeCommand.GetCertificateOutputFile(
+                    baseDirectory,
+                    certificateOutput,
+                    signingOutput,
+                    filesArgument);
 
                 int exitCode = await codeCommand.HandleAsync(
                     parseResult,
--- a/src/Sign.Cli/TrustedSigningCommand.cs
+++ b/src/Sign.Cli/TrustedSigningCommand.cs
@@ -105,7 +105,7 @@ namespace Sign.Cli
 
                 ArtifactSigningServiceProvider trustedSigningServiceProvider = new();
 
-                return codeCommand.HandleAsync(parseResult, serviceProviderFactory, trustedSigningServiceProvider, filesArgument);
+                return codeCommand.HandleAsync(parseResult, serviceProviderFactory, trustedSigningServiceProvider, filesArgument, cancellationToken);
             });
         }
     }
--- a/src/Sign.Cli/TrustedSigningCommand.cs
+++ b/src/Sign.Cli/TrustedSigningCommand.cs
@@ -113,19 +113,29 @@ namespace Sign.Cli
 
                 cancellationToken.ThrowIfCancellationRequested();
 
+                string? certificateOutput = parseResult.GetValue(CertificateOutputOption);
+                FileInfo? certificateOutputFile = null;
+
+                if (!string.IsNullOrEmpty(certificateOutput))
+                {
+                    DirectoryInfo baseDirectory = parseResult.GetValue(codeCommand.BaseDirectoryOption)!;
+                    certificateOutputFile = new FileInfo(CodeCommand.ExpandFilePath(baseDirectory, certificateOutput));
+
+                    if (certificateOutputFile.Exists)
+                    {
+                        throw new IOException($"The certificate output file already exists: '{certificateOutputFile.FullName}'.");
+                    }
+                }
+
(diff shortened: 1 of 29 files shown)


src/Sign.Cli/TrustedSigningCommand.cs
```diff
@@ -105,7 +105,7 @@ internal TrustedSigningCommand(CodeCommand codeCommand, IServiceProviderFactory
 
                 ArtifactSigningServiceProvider trustedSigningServiceProvider = new();
 
-                return codeCommand.HandleAsync(parseResult, serviceProviderFactory, trustedSigningServiceProvider, filesArgument);
+                return codeCommand.HandleAsync(parseResult, serviceProviderFactory, trustedSigningServiceProvider, filesArgument, cancellationToken);
```
dlemstra: I personally think it would make more sense to make it a specific option for this command and execute it after this handle command?

Jaxelr: +1 to this for the ergonomy and usability. Its more straightforward to have a command to retrieve the pubkey, although to be fair there's a small tiny percentage of possibility that the certificate might've rotated when the get is executed after the sign, this way we guarantee we use the same certificate on the operation.

jozefizso:
review-anvil-improve-pr completed on this PR. cc @jozefizso.

---

# ⚒️ review-anvil report

**Result:** All confirmed code-quality and correctness findings were fixed in five commits.
**Scope:** This PR adds provider-specific `--certificate-output` support for Artifact Signing and the obsolete Trusted Signing compatibility command.
**Verification:** 103 focused CLI tests passed. The macOS-compatible full run passed 327 of 330 tests; the three failures are unchanged Windows-path globbing tests. The normal Windows-target build remains unavailable on this macOS host because `mage.exe` is not installed.
**Checks:** 11 concerns checked; 6 confirmed, 4 ruled out, and 1 lowered in priority.
**Second check:** off; productive rounds applied fixes and reran the focused gate.

## Earlier review comments

- [Fixed] [Command-specific option and post-handler export](https://github.com/dotnet/sign/pull/1035#discussion_r3525225568) — `ArtifactSigningCommand` and `TrustedSigningCommand` now own the option and export only after their shared signing handler succeeds.

## What I noticed

- **RAV-RUN1-R1-F001 [high] file-system** — The original output write could replace a signed input or output with certificate bytes. Export now refuses existing destinations and uses `FileMode.CreateNew` as a race-safe final guard.
- **RAV-RUN1-R1-F002 [high] command ownership** — The option originally changed every code-signing provider. It now exists only on Artifact Signing and Trusted Signing.
- **RAV-RUN1-R3-F012 [high] workflow docs** — Folded PowerShell comments hid required Artifact Signing arguments. The command block no longer contains inline comments.
- **RAV-RUN1-R1-F003 [medium] cancellation** — Cancellation could leave signed artifacts without the requested certificate. The command now checks cancellation immediately before non-cancellable signing and completes post-sign export without observing later cancellation.
- **RAV-RUN1-R2-F008 [medium] workflow docs** — The example originally passed the provider-specific option to Azure Key Vault. It now uses Artifact Signing with endpoint, account, and profile arguments.
- **RAV-RUN1-R2-F009 [medium] partial success** — An existing certificate file was detected only after signing. Provider commands now validate it before signing.
- **RAV-RUN1-R2-F010 [medium] cancellation** — Cancellation during input preparation could still start signing. `CodeCommand` now checks the token again immediately before `SignAsync`.
- **RAV-RUN1-R3-F014 [medium] path validation** — Existing directories bypassed the file-only precheck. Both file and directory destinations are rejected before signing.
- **RAV-RUN1-R3-F015 [medium] output collision** — A new single-file signing output could match the certificate output. The command now rejects that collision before signing.

## Changes made / Things to try

- `b3847d7` — Scope certificate output to Artifact/Trusted Signing and preserve the exact cached provider.
- `99ba855` — Validate existing certificate output before signing and restore the final pre-sign cancellation check.
- `ceb942f` — Update the workflow to use Artifact Signing.
- `4f6b714` — Reject existing directories and single-file output collisions.
- `792cd19` — Keep every Artifact Signing argument active in folded PowerShell YAML.

## Set aside / Outside this change

- **RAV-RUN1-R1-F007 [nit] workflow docs** — The explicit step-output example is optional boilerplate, but it remains useful as an illustration for later workflow steps.
- **RAV-RUN1-R2-F011 [low] file-system** — Atomic temporary-file publication was set aside because create-new semantics already prevent destructive overwrite and the extra cleanup lifecycle is disproportionate here.
- **RAV-RUN1-R3-F016 [low] maintainability** — The two provider commands retain a small aligned orchestration block. The verifier lowered this to a suggestion because Trusted Signing is an obsolete compatibility command and no behavior currently differs.

<details>
<summary>Run details</summary>

- Target: `main...HEAD` on `feature/certificate_output_option` (26 files, +440/-14)
- Run ordinal: 1
- Rounds: 3 requested rounds completed; adaptive continuation stopped because the full local gate has pre-existing platform failures
- Mix: 2 codex-exec + 1 claude-exec
- Focus: correctness, maintainability, simplicity, production blast radius, code quality, and natural codebase fit
- Earlier review comments: 1 open comment; 1 fixed after checking
- Finding counts: 3 high, 6 medium, 2 low, 1 nit
- Checks: concerns=11; confirmed=6; ruled-out=4; set-aside=0; lowered=1
- Reviewer health: round 2 had one protocol failure after retry; round 3 recovered the same reviewer on its corrective retry
- Verification: focused tests passed; full local run had 3 unchanged macOS failures; normal Windows-target build lacked `mage.exe`

</details>

---
_Reviewed with [review-anvil](https://github.com/mrshu/agent-skills/#review-anvil)._


---

Started: 2026-08-10T20:58:04Z; Completed: 2026-08-10T21:44:29Z (outcome: success)

kartheekp-ms (changes requested):
~@jozefizso please rebase and update tests, since we migrated tests from [Moq to NSubstitute](https://github.com/dotnet/sign/pull/1054).~

Upon re-reviewing I noticed that this PR is for https://github.com/dotnet/sign/issues/1033 which is closed as duplicate of https://github.com/NuGet/NuGetGallery/issues/10027.

kartheekp-ms:
Upon re-reviewing I noticed that this PR is for https://github.com/dotnet/sign/issues/1033 which is closed as duplicate of https://github.com/NuGet/NuGetGallery/issues/10027.

Closing this PR.

PR #1047, opened 2026-07-27
Reject fragment delimiters in OPC part names

### LINKED ISSUE #998: Octothorpe symbol (#) in OPC part names is not handled properly
**Describe the bug**
If there is a content file containing octothorpe symbol (#) inside of an OPC package, sign CLI exits without any errors and outputs an invalid signed file. This seems to happen because this symbol denotes fragment URI component, which is not handled in any way when creating manifest references.

I attempted to fix this issue by modifying cli code to URL encode the '#' symbol during signing process, however this approach did not work since `System.IO.Packaging.PackageDigitalSignatureManager.VerifySignatures()` does not perform URL-decoding while resolving OPC references.

I wonder about how this symbol could be handled in OPC file formats in general. I know that at least a part of them simply do not allow this symbol. However, on the other hand - octothorpe is a completely valid symbol in Windows filesystem.

**Repro steps**
- Create a vsix package, which includes a file named ab#c.txt with a test string inside.
- Sign the file using the sign cli (and see that it exits without any errors).
- Run powershell script to validate the signature and see that it fails:
```powershell
Add-Type -Assembly "WindowsBase"
$Package = [System.IO.Packaging.Package]::Open("test.vsix", [System.IO.FileMode]::Open, [System.IO.FileAccess]::Read)
[System.IO.Packaging.PackageDigitalSignatureManager]::New($Package).VerifySignatures($false)
```

**Expected behavior**
When signing an OPC format (for example: *.vsix) file containing a content file with '#' in its name - it is either encoded, or an error is returned.

**Actual behavior**
Sign CLI exits without errors and results in an invalid signed file.

**Additional context**
- dotnet/sign version: 0.9.1-beta.26127.1+69657bd184d04b23782585fe0f85a510663142ac

Guflly:
Fixes #998.

Rejects raw `#` characters in OPC part names before signing instead of producing an invalid signature. The tests cover both root and nested entries and verify that no signature is left behind.

This follows the approach in #1042, which was closed before review.

`eng/common/Build.ps1 -restore -build -test -sign -pack -publish -ci` passes.

### DIFF
--- a/src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
+++ b/src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
@@ -20,7 +20,7 @@ namespace Sign.Core
         {
             if (path.Contains('#'))
             {
-                throw new InvalidDataException($"Package part name '{path}' contains the unsupported '#' character.");
+                throw new InvalidDataException(string.Format(Resources.VSIXSignToolOpcPartNameContainsFragmentDelimiter, path));
             }
 
             Uri = new Uri(OpcPackage.BasePackageUri, path);
--- a/src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
+++ b/src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
@@ -18,6 +18,11 @@ namespace Sign.Core
 
         internal OpcPart(OpcPackage package, string path, ZipArchiveEntry entry, OpcPackageFileMode mode)
         {
+            if (path.Contains('#'))
+            {
+                throw new InvalidDataException($"Package part name '{path}' contains the unsupported '#' character.");
+            }
+
             Uri = new Uri(OpcPackage.BasePackageUri, path);
             Package = package;
             _path = path;
--- a/src/Sign.Core/Resources.Designer.cs
+++ b/src/Sign.Core/Resources.Designer.cs
@@ -356,6 +356,15 @@ namespace Sign.Core {
                 return ResourceManager.GetString("VSIXSignToolOpcContentTypeInvalid", resourceCulture);
             }
         }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Package part name &apos;{0}&apos; contains the unsupported &apos;#&apos; character..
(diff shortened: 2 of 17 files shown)


Guflly:
@dotnet-policy-service agree

src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
```diff
@@ -18,6 +18,11 @@ internal sealed class OpcPart : IEquatable<OpcPart>
 
         internal OpcPart(OpcPackage package, string path, ZipArchiveEntry entry, OpcPackageFileMode mode)
         {
+            if (path.Contains('#'))
+            {
+                throw new InvalidDataException($"Package part name '{path}' contains the unsupported '#' character.");
```
dlemstra: This should probably come from the resources like we do with the other messages.

src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
```diff
@@ -111,4 +116,4 @@ private OpcRelationships ConstructRelationships()
         /// <inheritdoc />
         public override int GetHashCode() => Uri.GetHashCode();
     }
-}
```
dlemstra, line 114: Please keep this in the file.

src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
Guflly: Moved it into the resources.

src/Sign.Core/Tools/VsixSignTool/OpcPart.cs
Guflly: Restored it.

kartheekp-ms (changes requested):
@Guflly please rebase since we migrated tests from [Moq to NSubstitute](https://github.com/dotnet/sign/pull/1054).

dtivel:
Fixed via https://github.com/dotnet/sign/commit/b8c81ed9011aa1452baab316d30b47902122e1de

Issue #1049, opened 2026-08-08
Implement ClickOnce signing algorithm version 2


dtivel:
## Summary

Implement the [ClickOnce signing algorithm version 2 specification](https://github.com/dotnet/sign/blob/main/docs/specs/ClickOnce-Signing-Algorithm.md) through small PRs that merge serially without exposing incomplete behavior.

## Planned PRs

- [ ] **PR 1: Manifest foundation**: Add packaged MSBuild dependencies and dormant typed manifest APIs. Actual PR: #1050
- [ ] **PR 2: File graph**: Add manifest-driven dependency discovery and file classification. Actual PR: #1051
- [ ] **PR 3: Safe staging**: Add contained staging, collision checks, mapped-suffix handling, and reference rebinding. Actual PR: #1058
- [ ] **PR 4: Signing operation coordinator**: Add invocation-scoped coordination and stable signing-source identities.
- [ ] **PR 5: Signing lifecycle**: Add the shared staging, signing, copy-back, and output lifecycle without activating coordination.
- [ ] **PR 6: Coordinator integration**: Activate invocation-scoped coordination across signing formats and containers.
- [ ] **PR 7: Version 2 application manifests**: Add dormant version-2 application-manifest signing.
- [ ] **PR 8: Version 2 deployment manifests**: Add the dormant deployment-manifest pipeline and ordering behavior.
- [ ] **PR 9: CLI exposure and package validation**: Expose version 2 and prove extracted `sign.dll` uses packaged MSBuild dependencies.
- [ ] **PR 10: User documentation**: Document options, rollout, re-signing modes, examples, and operational behavior.

## Completion criteria

- The specification is fully implemented and documented.
- Every PR includes its required tests and can merge in order.
- Version 1 remains the default during the initial rollout.
- Version 2 is not release-ready before PR 10.

PR #1050, opened 2026-08-08
Add ClickOnce manifest foundation

dtivel:
Part of #1049.

## Summary

Add the dormant manifest foundation required by the [ClickOnce signing algorithm version 2 specification](https://github.com/dotnet/sign/blob/main/docs/specs/ClickOnce-Signing-Algorithm.md).

- Add typed application and deployment manifest abstractions, adapters, and readers.
- Extract shared manifest behavior into `ClickOnceManifestAdapter`.
- Use `IReadOnlyList` for resolution inputs and `FileInfo` for serialization destinations.
- Fix metadata updates and serialization to use ManifestUtilities' `v4.5` behavior, preserving SHA-256 digest values and declarations.
- Always preserve the source manifest XML so elements and attributes not modeled by ManifestUtilities survive serialization.
- Remove stale `publisherIdentity` and XML signature elements from the preserved XML before rewriting so regenerated signing metadata remains authoritative.
- Read preserved XML with DTD processing prohibited and no XML resolver.
- Redistribute the required MSBuild manifest runtime assemblies.
- Verify that those assemblies are included in the Sign CLI package.
- Add coverage for typed reads, manifest rejection, stream reuse, adapter contracts, metadata updates, application/deployment serialization, repeated writes, and preservation of unmodeled VSTO XML.

The preservation behavior is format-neutral. This PR does not classify manifests as ClickOnce or VSTO and adds no technology-specific API.

## Behavior

The new implementation is dormant:

- It is not registered with dependency injection.
- It is not reachable from production signing.
- It adds no CLI options or dispatch changes.
- Existing ClickOnce version 1 behavior remains unchanged.

## Testing

- Focused ClickOnce manifest tests: 33 passed, 0 failed
- Coverage verifies exact SHA-256 digest values and serialized SHA-256 `DigestMethod` declarations.
- Preservation coverage verifies that unmodeled XML survives rewriting while stale publisher and signature metadata does not.


src/Sign.Core/Tools/ClickOnce/ApplicationManifestAdapter.cs
```diff
@@ -0,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+using Microsoft.Build.Tasks.Deployment.ManifestUtilities;
+
+namespace Sign.Core
```
dlemstra, line 7: Wondering if new files should have file scope namespaces? And then at some point change the rest to file scope also?

src/Sign.Core/Tools/ClickOnce/ClickOnceManifestReaderProvider.cs
```diff
@@ -0,0 +1,28 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+namespace Sign.Core
+{
+    internal sealed class ClickOnceManifestReaderProvider
+    {
+        private readonly Lazy<IClickOnceManifestReader> _reader;
```
dlemstra: Why the lazy? I don't understand why you would need that "complexity". Or is that for future changes?

src/Sign.Core/Tools/ClickOnce/ApplicationManifestAdapter.cs
dtivel: I kept the existing repository convention. There currently aren’t any file-scoped namespaces in the codebase, so I think converting to them would be better handled as a separate repository-wide style change.

src/Sign.Core/Tools/ClickOnce/ClickOnceManifestReaderProvider.cs
dtivel: I agree. Looking ahead to the eventual wiring, `CodeCommand` knows the effective version before the service provider is built, so PR 9 can register `IClickOnceManifestReader` only for version 2. Microsoft DI already constructs singletons lazily, making the custom `Lazy<T>` provider redundant. I've removed the provider and its tests.

src/Sign.Core/Tools/ClickOnce/ApplicationManifestAdapter.cs
```diff
@@ -0,0 +1,53 @@
...
+            ArgumentException.ThrowIfNullOrWhiteSpace(targetFrameworkVersion, nameof(targetFrameworkVersion));
+
+            _manifest.UpdateFileInfo(targetFrameworkVersion);
+        }
+
+        public void Write(Stream stream)
+        {
+            ArgumentNullException.ThrowIfNull(stream, nameof(stream));
+
+            ManifestWriter.WriteManifest(_manifest, stream);
```
kartheekp-ms: ManifestWriter.WriteManifest(_manifest, stream)  serializes referenced-file hashes as SHA-1, even when `UpdateFileInfo("v4.5")` generated SHA-256 digest values. This produces a manifest whose 32-byte digest value conflicts with its declared  DigestMethod .

MSBuild’s stream overload passes  null  as the target framework version:
https://github.com/dotnet/msbuild/blob/f08c80626816028e6197636edb41e918c530a632/src/Tasks/ManifestUtil/ManifestWriter.cs#L92-L95

A null framework version selects  write2.xsl :
https://github.com/dotnet/msbuild/blob/f08c80626816028e6197636edb41e918c530a632/src/Tasks/ManifestUtil/ManifestWriter.cs#L118-L125

 write2.xsl  declares SHA-1:
https://github.com/dotnet/msbuild/blob/f08c80626816028e6197636edb41e918c530a632/src/Tasks/ManifestUtil/write2.xsl#L344

By comparison,  write3.xsl  declares SHA-256:
https://github.com/dotnet/msbuild/blob/f08c80626816028e6197636edb41e918c530a632/src/Tasks/ManifestUtil/write3.xsl#L344

The [ClickOnce signing specification](https://github.com/dotnet/sign/blob/main/docs/specs/ClickOnce-Signing-Algorithm.md#appendix-b--signing-algorithm-version-2) requires  UpdateFileInfo("v4.5") specifically to retain SHA-256 behavior:

Could we change this adapter to use the framework-aware path overload?

ManifestWriter.WriteManifest(_manifest, path, "v4.5");

I leveraged Copilot to write a test for this scenario, and it failed.

```cs
[Fact]
public void ApplicationManifestAdapter_Write_PreservesSha256DigestMethod()
{
    const string TargetFrameworkVersion = "v4.5";
    const string Sha256Algorithm =
        "http://www.w3.org/2000/09/xmldsig#sha256";
    XNamespace dsig =
        "http://www.w3.org/2000/09/xmldsig#";

    string temporaryFilePath = Path.GetTempFileName();

    try
    {
        File.WriteAllText(temporaryFilePath, "payload");

        ApplicationManifest manifest = CreateApplicationManifest();
        manifest.FileReferences.Add(new FileReference(temporaryFilePath));

        ApplicationManifestAdapter adapter = new(manifest);
        adapter.ResolveFiles(
            new[] { Path.GetDirectoryName(temporaryFilePath)! });
        adapter.UpdateFileInfo(TargetFrameworkVersion);

        using MemoryStream stream = new();
        adapter.Write(stream);
        stream.Position = 0;

        XDocument document = XDocument.Load(stream);
        XElement digestMethod = Assert.Single(
            document.Descendants(dsig + "DigestMethod"));

        Assert.Equal(
            Sha256Algorithm,
            digestMethod.Attribute("Algorithm")?.Value);
    }
    finally
    {
        File.Delete(temporaryFilePath);
    }
```

src/Sign.Core/Tools/ClickOnce/DeployManifestAdapter.cs
```diff
@@ -0,0 +1,53 @@
...
+            ArgumentException.ThrowIfNullOrWhiteSpace(targetFrameworkVersion, nameof(targetFrameworkVersion));
+
+            _manifest.UpdateFileInfo(targetFrameworkVersion);
+        }
+
+        public void Write(Stream stream)
+        {
+            ArgumentNullException.ThrowIfNull(stream, nameof(stream));
+
+            ManifestWriter.WriteManifest(_manifest, stream);
```
kartheekp-ms: Same as https://github.com/dotnet/sign/pull/1050/changes#r3753406000

src/Sign.Core/Tools/ClickOnce/ApplicationManifestAdapter.cs
```diff
@@ -0,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+using Microsoft.Build.Tasks.Deployment.ManifestUtilities;
+
+namespace Sign.Core
+{
+    internal sealed class ApplicationManifestAdapter : IApplicationManifest
```
kartheekp-ms: I see code duplication in `ApplicationManifestAdapter` and `DeployManifestAdapter` classes. I think we extract the common logic in a base class `ClickOnceManifestAdapter` and move the common logic there.

src/Sign.Core/Tools/ClickOnce/ApplicationManifestAdapter.cs
dtivel: Fixed in e0f67a1. The adapter now writes through the framework-aware path overload with `v4.5`, and `UpdateFileInfo()` also fixes the version internally. Regression coverage verifies both the exact SHA-256 digest and serialized SHA-256 `DigestMethod`.

src/Sign.Core/Tools/ClickOnce/ApplicationManifestAdapter.cs
dtivel: Fixed in e0f67a1. I extracted `ClickOnceManifestAdapter<TManifest>` for the shared identity, references, entry point, diagnostics, writable state, resolution, metadata update, and serialization behavior. The concrete adapters now contain only their type-specific properties.

src/Sign.Core/Tools/ClickOnce/DeployManifestAdapter.cs
dtivel: Fixed in e0f67a1 through the shared adapter implementation. The deployment regression test verifies the exact SHA-256 entry-point digest and serialized SHA-256 `DigestMethod`.

src/Sign.Core/Tools/ClickOnce/ClickOnceManifestReader.cs
```diff
@@ -0,0 +1,215 @@
...
+
+namespace Sign.Core
+{
+    internal sealed class ClickOnceManifestReader : IClickOnceManifestReader
+    {
+        private const string AssemblyV2Namespace =
+            "urn:schemas-microsoft-com:asm.v2";
+        private const string SignatureNamespace =
+            "http://www.w3.org/2000/09/xmldsig#";
+        public bool TryReadApplicationManifest(
```
kartheekp-ms, line 19: nit: a space here will be helpful.

src/Sign.Core/Tools/ClickOnce/ClickOnceManifestReader.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                throw new ArgumentException(
+                    "The manifest stream must be positioned at the beginning.",
+                    nameof(stream));
+            }
+
+            return ManifestReader.ReadManifest(stream, preserveStream: true);
+        }
+
+        private static Stream SanitizePreservedInput(Stream input)
```
kartheekp-ms: Couuld we simplify this by loading an `XmlDocument` with `PreserveWhitespace = true`, removing the stale root-level elements, and saving it to a `MemoryStream`? That would preserve the current behavior while avoiding the custom reader-advancement state and node-copying switch. It may also help to separate stream transformation from ownership/disposal so each method’s contract is clearer. I tried this approach locally, and all 33 focused ClickOnce tests passed.

```cs
private static Stream SanitizePreservedInput(Stream input)
{
    input.Position = 0;
    MemoryStream output = new();

    XmlReaderSettings readerSettings = new()
    {
        DtdProcessing = DtdProcessing.Prohibit,
        XmlResolver = null
    };

    XmlWriterSettings writerSettings = new()
    {
        CloseOutput = false,
        Encoding = new UTF8Encoding(
            encoderShouldEmitUTF8Identifier: false),
        Indent = false
    };

    try
    {
        XmlDocument document = new()
        {
            PreserveWhitespace = true,
            XmlResolver = null
        };

        using (XmlReader reader = XmlReader.Create(input, readerSettings))
        {
            document.Load(reader);
        }

        XmlElement root = document.DocumentElement
            ?? throw new XmlException(
                "The manifest does not contain a document element.");

        // Traverse backwards because removing a child changes the collection.
        for (int i = root.ChildNodes.Count - 1; i >= 0; --i)
        {
            if (root.ChildNodes[i] is XmlElement element &&
                IsStaleSigningElement(element))
            {
                root.RemoveChild(element);
            }
        }

        using (XmlWriter writer = XmlWriter.Create(output, writerSettings))
        {
            document.Save(writer);
        }

        output.Position = 0;

        return output;
    }
    catch
    {
        output.Dispose();
        throw;
    }
    finally
    {
        input.Dispose();
    }
}

private static bool IsStaleSigningElement(XmlElement element)
{
    bool isPublisherIdentity =
        element.LocalName == "publisherIdentity" &&
        element.NamespaceURI == AssemblyV2Namespace;

    bool isSignature =
        element.LocalName == "Signature" &&
        element.NamespaceURI == SignatureNamespace;

    return isPublisherIdentity || isSignature;
}
```

src/Sign.Core/Tools/ClickOnce/IApplicationManifest.cs
```diff
@@ -0,0 +1,15 @@
...
+// See the LICENSE.txt file in the project root for more information.
+
+using Microsoft.Build.Tasks.Deployment.ManifestUtilities;
+
+namespace Sign.Core
+{
+    internal interface IApplicationManifest : IClickOnceManifest
+    {
+        AssemblyReferenceCollection AssemblyReferences { get; }
+        AssemblyReference? EntryPoint { get; }
```
kartheekp-ms: `AssemblyReferences` and `EntryPoint` are duplicated in application manifest and deployment manifest interfaces. Can we move these properties to the IClickOnceManifest interface?

kartheekp-ms (approved):

PR #1051, opened 2026-08-09
Add ClickOnce publish-layout resolution

dtivel:
Part of #1049.

## Summary

- Add dormant ClickOnce publish-layout resolution through:
  - `ClickOnceApplicationPublishLayoutResolver`
  - `ClickOnceDeploymentPublishLayoutResolver`
  - `ClickOncePayloadResolver`
- Represent the resolved layout with strongly typed models:
  - `ResolvedClickOncePublishLayout`
  - `ResolvedClickOnceDeployment`
  - `ResolvedClickOnceApplication`
  - `ResolvedClickOncePayload`
  - `ResolvedClickOnceAdjacentExecutable`
- Report resolution failures through `ClickOncePublishLayoutResolutionException` while preserving ordered manifest diagnostics.
- Read deployment and application manifests through the preservation-safe typed manifest-reader API for every resolution read.
- Require a deployment manifest model to expose exactly one assembly reference, no file references, and an entry point that is the same non-prerequisite, non-optional reference. Inspect the preserved deployment XML to reject resource-fallback attributes that `ManifestUtilities` neither models nor preserves. Reject unsupported shapes before file resolution so `ManifestUtilities` or `mage.exe` cannot reorder or rewrite them before signing.
- Reject rooted deployment-entry-point and application-payload target paths before resolution because they cannot be represented beneath a staging root. Preserve relative target paths, including parent traversal, for containment validation against the eventual staging root.
- Resolve the deployment manifest's referenced application manifest, then resolve physical application payloads using the specified mapping and directory-fallback rules.
- Preserve each payload's source and manifest reference while snapshotting its target path. `ResolvedClickOncePayload` validates that the source filename matches the target filename, with or without exactly one additional `.deploy` suffix, and derives `IsFileExtensionMapped` from that relationship.
- Model adjacent executables separately with a constrained setup/launcher kind. `ResolvedClickOnceAdjacentExecutable` derives its target path from its source filename and validates the filename against its kind.
- Discover only applicable adjacent `setup.exe` and unreferenced root-level `Launcher.exe` files; referenced launchers remain application payloads.
- Use neutral "deployment manifest" and "application manifest" terminology in user-facing resolution errors.
- Add coverage for deployment-reference validation, required and optional physical references, multiple versions, malformed and wrong-type manifests, mapping, fallback and diagnostic ordering, adjacent executables, VSTO layouts, target-path validation and preservation, and model invariants.

## Intentional scope boundary

This PR resolves and models the publish layout only. It does not register the resolvers with dependency injection and does not change production signing behavior or the CLI surface. Diagnostic text may reference the planned `--clickonce-signing-version 2 --no-update-clickonce-manifest` option, but the resolvers remain dormant until signer integration adds that option.

Staging, relative-path containment validation, copying, temporary `.deploy` suffix removal and restoration, manifest metadata updates, signing-operation coordination, and signing remain deferred to the signer-integration work. Resolution rejects rooted target paths but does not rewrite or remap accepted relative paths. Staging must ensure those relative paths remain within the staging directory and fail when a referenced layout cannot be represented safely.

The deployment resolver establishes that `Application.Source` is the file initially identified by `Deployment.ApplicationManifestReference.ResolvedPath`. This is a resolution-time postcondition rather than a durable model invariant because staging must redirect that mutable `ResolvedPath`. The staging implementation must stage `Application.Source`, redirect the deployment entry point to that exact staged copy, verify that it is contained within the staging directory, and only then update deployment-manifest metadata. At the start of staging, it must capture the deployment entry point's target path. Immediately before deployment-manifest metadata update and serialization, it must fail unless the live entry point is still the resolved `ApplicationManifestReference`, remains the sole assembly reference with no file references, and its target path still ordinally equals the captured value. It must not reconcile or serialize mutated deployment structure.

For payloads, `ResolvedClickOncePayload.TargetPath` is an immutable resolution-time snapshot of `Reference.TargetPath`, not a live projection. Staging must use that snapshot to determine the intended layout, preserve the manifest-relative meaning rather than remapping it, and bind each mutable reference's `ResolvedPath` to the corresponding contained staged file before updating metadata. Before computing staging destinations or collision claims, and again immediately before updating application-manifest metadata and serializing the manifest, staging must verify with ordinal comparison that each mutable `Reference.TargetPath` still equals its resolved snapshot. A mismatch must fail staging rather than reconcile or emit an inconsistent manifest.

The deployment-directory payload fallback is a source-discovery mechanism. Signer integration must retain an explicit mapping from every staged file to the original `ResolvedClickOncePayload.Source` it was staged from and copy signed bytes back to that exact source path. It must not derive copy-back destinations from staging-relative paths or reuse a directory-mirroring copy-back seam that would create a new application-directory file for a deployment-directory fallback source. Integration coverage must verify that a fallback payload is returned to its original source and that no unintended target-relative file is created.

Discovery can also produce two resolved references to one physical file: a root-level `setup.exe` referenced by the application manifest is classified as both a `ResolvedClickOncePayload` and a `ResolvedClickOnceAdjacentExecutable`, with both `Source` values identifying the same file and potentially the same target path. Signer integration must apply the same source-keyed coordinated signing operation to both references so the file is staged, signed, and copied back exactly once rather than processed as two independent operations.

### DIFF
--- a/src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraph.cs
+++ b/src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraph.cs
@@ -1,0 +1,41 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+namespace Sign.Core
+{
+    internal sealed class ClickOnceFileGraph
+    {
+        internal ClickOnceFileGraph(
+            ClickOnceFileGraphEntry? deploymentManifest,
+            IDeployManifest? deployManifest,
+            ClickOnceFileGraphEntry applicationManifest,
+            IApplicationManifest applicationManifestModel,
+            IEnumerable<ClickOnceFileGraphEntry> payloads,
+            IEnumerable<ClickOnceFileGraphEntry> adjacentExecutables,
+            IEnumerable<ClickOnceManifestDiagnostic> diagnostics)
+        {
+            ArgumentNullException.ThrowIfNull(applicationManifest, nameof(applicationManifest));
+            ArgumentNullException.ThrowIfNull(applicationManifestModel, nameof(applicationManifestModel));
+            ArgumentNullException.ThrowIfNull(payloads, nameof(payloads));
+            ArgumentNullException.ThrowIfNull(adjacentExecutables, nameof(adjacentExecutables));
+            ArgumentNullException.ThrowIfNull(diagnostics, nameof(diagnostics));
+
+            DeploymentManifest = deploymentManifest;
+            DeployManifest = deployManifest;
+            ApplicationManifest = applicationManifest;
+            ApplicationManifestModel = applicationManifestModel;
+            Payloads = payloads.ToArray();
+            AdjacentExecutables = adjacentExecutables.ToArray();
+            Diagnostics = diagnostics.ToArray();
+        }
+
+        internal ClickOnceFileGraphEntry? DeploymentManifest { get; }
+        internal IDeployManifest? DeployManifest { get; }
+        internal ClickOnceFileGraphEntry ApplicationManifest { get; }
+        internal IApplicationManifest ApplicationManifestModel { get; }
+        internal IReadOnlyList<ClickOnceFileGraphEntry> Payloads { get; }
+        internal IReadOnlyList<ClickOnceFileGraphEntry> AdjacentExecutables { get; }
+        internal IReadOnlyList<ClickOnceManifestDiagnostic> Diagnostics { get; }
+    }
+}
--- a/src/Sign.Core/Resources.Designer.cs
+++ b/src/Sign.Core/Resources.Designer.cs
@@ -97,6 +97,159 @@ namespace Sign.Core {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Application manifest &apos;{0}&apos; contains invalid target path &apos;{1}&apos;..
+        /// </summary>
+        internal static string ClickOnceApplicationManifestInvalidTargetPath {
+            get {
+                return ResourceManager.GetString("ClickOnceApplicationManifestInvalidTargetPath", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Application manifest &apos;{0}&apos; contains a required file reference without a target path..
+        /// </summary>
+        internal static string ClickOnceApplicationManifestMissingTargetPath {
+            get {
+                return ResourceManager.GetString("ClickOnceApplicationManifestMissingTargetPath", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to access source candidate &apos;{2}&apos; for target path &apos;{1}&apos; in application manifest &apos;{0}&apos;..
+        /// </summary>
+        internal static string ClickOnceApplicationManifestPayloadProbeFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceApplicationManifestPayloadProbeFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to read application manifest &apos;{0}&apos;..
+        /// </summary>
+        internal static string ClickOnceApplicationManifestReadFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceApplicationManifestReadFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Application manifest &apos;{0}&apos; references required file &apos;{1}&apos;, but it could not be found..
+        /// </summary>
+        internal static string ClickOnceApplicationManifestRequiredFileNotFound {
+            get {
+                return ResourceManager.GetString("ClickOnceApplicationManifestRequiredFileNotFound", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to resolve files referenced by application manifest &apos;{0}&apos;..
+        /// </summary>
+        internal static string ClickOnceApplicationManifestResolveFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceApplicationManifestResolveFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Deployment manifest &apos;{0}&apos; references application manifest &apos;{1}&apos;, but the file does not exist..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestApplicationManifestNotFound {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestApplicationManifestNotFound", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to access launcher candidate &apos;{0}&apos; associated with deployment manifest &apos;{1}&apos;..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestLauncherProbeFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestLauncherProbeFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Deployment manifest &apos;{0}&apos; does not identify an application manifest..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestMissingEntryPoint {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestMissingEntryPoint", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to read deployment manifest &apos;{0}&apos;..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestReadFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestReadFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to read application manifest &apos;{0}&apos; referenced by deployment manifest &apos;{1}&apos;..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestReferencedApplicationReadFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestReferencedApplicationReadFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to File &apos;{0}&apos; referenced by deployment manifest &apos;{1}&apos; is not an application manifest..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestReferencedApplicationWrongType {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestReferencedApplicationWrongType", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to resolve files referenced by deployment manifest &apos;{0}&apos;..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestResolveFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestResolveFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to access setup bootstrapper candidate &apos;{0}&apos; associated with deployment manifest &apos;{1}&apos;..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestSetupProbeFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestSetupProbeFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Deployment manifest &apos;{0}&apos; did not resolve its referenced application manifest &apos;{1}&apos;..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestUnresolvedApplicationManifest {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestUnresolvedApplicationManifest", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Deployment manifest &apos;{0}&apos; must contain exactly one non-prerequisite application manifest reference and no file references..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestUnsupportedReferences {
+            get {
+                return ResourceManager.GetString("ClickOnceDeploymentManifestUnsupportedReferences", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to File &apos;{0}&apos; is not a deployment manifest..
+        /// </summary>
+        internal static string ClickOnceDeploymentManifestWrongType {
+            get {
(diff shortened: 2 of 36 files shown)


kartheekp-ms (dismissed):

src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraph.cs
```diff
@@ -0,0 +1,41 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+namespace Sign.Core
+{
+    internal sealed class ClickOnceFileGraph
+    {
+        internal ClickOnceFileGraph(
+            ClickOnceFileGraphEntry? deploymentManifest,
+            IDeployManifest? deployManifest,
```
kartheekp-ms, line 11: Could we strengthen the file-graph model before merging this PR? I think follow-up PR #1058 demonstrates that the current shape requires consumers to reconstruct invariants that could instead be encoded here.

## 1. Deployment state can be half-present

[`ClickOnceFileGraph`](https://github.com/dotnet/sign/blob/57f55605ecd907af89825161696de4fd92cf9dfe/src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraph.cs#L9-L38) represents deployment input with two independently nullable properties:

```csharp
ClickOnceFileGraphEntry? DeploymentManifest
IDeployManifest? DeployManifest
```

That produces four possible states:

| `DeploymentManifest` | `DeployManifest` | Meaning |
|---|---|---|
| present | present | Valid deployment input |
| absent | absent | Valid explicit application input |
| present | absent | Invalid half-present deployment |
| absent | present | Invalid half-present deployment |

The constructor currently accepts all four. For example, this compiles:

```csharp
ClickOnceFileGraph graph = new(
    deploymentManifest: deploymentEntry,
    deployManifest: null,
    applicationManifest,
    applicationManifestModel,
    payloads,
    adjacentExecutables,
    diagnostics);
```

The inverse also compiles:

```csharp
ClickOnceFileGraph graph = new(
    deploymentManifest: null,
    deployManifest,
    applicationManifest,
    applicationManifestModel,
    payloads,
    adjacentExecutables,
    diagnostics);
```

This is already visible in [PR #1058's candidate selection](https://github.com/dotnet/sign/blob/bddf243e6bad4917453e7fd977ef9d158de8c4f4/src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraphStager.cs#L73-L113):

```csharp
bool isDeploymentInput = graph.DeploymentManifest is not null;

if (isDeploymentInput)
{
    candidates.Add(
        new StagingCandidate(
            graph.DeploymentManifest!,
            isUpdateInputOnly: false));
}
```

The stager must choose one nullable property as the authority and use `!`. Consequently:

- `DeploymentManifest != null` and `DeployManifest == null` is silently treated as deployment input.
- `DeploymentManifest == null` and `DeployManifest != null` is silently treated as application input.

The current resolvers construct valid pairs, but the result type itself does not preserve that invariant for future consumers.

## 2. Deployment-only adjacent files can exist without a deployment

The model also permits this:

```csharp
ClickOnceFileGraph graph = new(
    deploymentManifest: null,
    deployManifest: null,
    applicationManifest,
    applicationManifestModel,
    payloads: Array.Empty<ClickOnceFileGraphEntry>(),
    adjacentExecutables: new[]
    {
        new ClickOnceFileGraphEntry(
            setupFile,
            "setup.exe",
            ClickOnceFileGraphEntryKind.Setup)
    },
    diagnostics);
```

This represents an explicit application-manifest input containing a deployment-only `setup.exe`.

PR #1058 will actually include that file in default staging because [adjacent executables are added independently of `isDeploymentInput`](https://github.com/dotnet/sign/blob/bddf243e6bad4917453e7fd977ef9d158de8c4f4/src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraphStager.cs#L115-L133):

```csharp
if (mode == ClickOnceSigningMode.Default)
{
    candidates.AddRange(
        graph.AdjacentExecutables.Select(
            entry => new StagingCandidate(
                entry,
                isUpdateInputOnly: false)));
}
```

The resolvers currently avoid constructing this state, but the aggregate allows it and the next layer assigns it behavior.

## 3. A generic entry permits contradictory roles

[`ClickOnceFileGraphEntry`](https://github.com/dotnet/sign/blob/57f55605ecd907af89825161696de4fd92cf9dfe/src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraphEntry.cs#L11-L32) accepts any combination of:

```csharp
Source
TargetPath
Kind
ManifestReference?
MappingAddedSuffix?
```

For example, this payload can be mislabeled as a setup executable:

```csharp
ClickOnceFileGraphEntry entry = new(
    payloadFile,
    @"bin\payload.dll",
    ClickOnceFileGraphEntryKind.Setup,
    payloadReference);
```

PR #1058 [uses `Kind` to determine the target-path base directory](https://github.com/dotnet/sign/blob/bddf243e6bad4917453e7fd977ef9d158de8c4f4/src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraphStager.cs#L595-L614):

```csharp
string basePath =
    graphEntry.Kind == ClickOnceFileGraphEntryKind.Payload
        ? applicationDirectoryPath
        : _rootPath;
```

The mislabeled payload would therefore be staged relative to the deployment root rather than the application-manifest directory.

Other currently constructible combinations include:

```csharp
// Payload without the manifest reference that owns it.
new ClickOnceFileGraphEntry(
    payloadFile,
    "payload.dll",
    ClickOnceFileGraphEntryKind.Payload,
    manifestReference: null);

// Setup executable with a manifest reference and mapping suffix.
new ClickOnceFileGraphEntry(
    setupFile,
    "setup.exe",
    ClickOnceFileGraphEntryKind.Setup,
    payloadReference,
    mappingAddedSuffix: ".deploy");

// Entry path disagrees with the manifest reference path.
new ClickOnceFileGraphEntry(
    payloadFile,
    @"bin\payload.dll",
    ClickOnceFileGraphEntryKind.Payload,
    new FileReference
    {
        TargetPath = @"lib\payload.dll"
    });
```

The follow-up therefore needs additional validation and null-forgiving operations, such as [validating that a mapped entry has a reference](https://github.com/dotnet/sign/blob/bddf243e6bad4917453e7fd977ef9d158de8c4f4/src/Sign.Core/Tools/ClickOnce/ClickOnceFileGraphStager.cs#L353-L380) and later using [`ManifestReference!`](https://github.com/dotnet/sign/blob/bddf243e6bad4917453e7fd977ef9d158de8c4f4/src/Sign.Core/Tools/ClickOnce/ClickOnceStagingResult.cs#L270-L283).

## Suggested model

Would it be safer to represent the fixed hierarchy directly?

```csharp
internal sealed class ResolvedClickOncePublication
{
    internal ResolvedClickOnceDeployment? Deployment { get; }
    internal ResolvedClickOnceApplication Application { get; }
    internal IReadOnlyList<ClickOnceManifestDiagnostic> Diagnostics { get; }
}

internal sealed class ResolvedClickOnceDeployment
{
    internal FileInfo Source { get; }
    internal IDeployManifest Manifest { get; }
    internal AssemblyReference ApplicationManifestReference { get; }
    internal IReadOnlyList<ClickOnceAdjacentExecutable> AdjacentExecutables { get; }
}

internal sealed class ResolvedClickOnceApplication
{
    internal FileInfo Source { get; }
    internal IApplicationManifest Manifest { get; }
    internal IReadOnlyList<ResolvedClickOncePayload> Payloads { get; }
}

internal sealed class ResolvedClickOncePayload
{
    internal FileInfo Source { get; }
    internal BaseReference Reference { get; }
    internal string ApplicationRelativeTargetPath => Reference.TargetPath;
    internal string? MappingAddedSuffix { get; }
}
```

This would make the important invariants structural:

- deployment entry and model are either both present or both absent;
- adjacent executables cannot exist without a deployment;
- every payload has a manifest reference;
- payload paths are explicitly application-relative;
- the exposed target path cannot disagree with `Reference.TargetPath`;
- downstream code does not need `Kind` branching to recover an object's role.

dtivel: Thanks for the detailed feedback, @kartheekp-ms . I adopted the typed hierarchy you suggested and addressed the three invalid model states you identified:

- `ResolvedClickOncePublishLayout` has one optional `ResolvedClickOnceDeployment` and a required `ResolvedClickOnceApplication`, eliminating independently nullable deployment state.
- Deployment groups its source, manifest, and application-manifest reference. The constructor requires that reference to be the manifest's actual `EntryPoint` object.
- Application groups its source, manifest, and typed payloads.
- Dedicated payload and adjacent-executable types replace the generic graph entry and broad role enum. Every payload requires its manifest reference, while adjacent executables cannot exist without deployment context.
- Payloads validate the source/target filename relationship and derive `IsFileExtensionMapped`; adjacent executables derive their target path from their source filename and validate `setup.exe` or `Launcher.exe` against their narrow kind.

I made a few deliberate departures from the suggested shape:

- I used `ResolvedClickOncePublishLayout` rather than `ResolvedClickOncePublication`, because the result describes the resolved on-disk layout.
- Adjacent executables remain on the top-level layout rather than under `Deployment`. They are discovered by publish-directory convention rather than referenced by the deployment manifest, but the layout constructor still requires deployment context.
- `ResolvedClickOncePayload.TargetPath` is an immutable resolution-time snapshot of `Reference.TargetPath`, not a live projection. This preserves the target value against which source matching and file-extension mapping were validated, while the underlying `ManifestUtilities` reference remains mutable.
- I retained `TargetPath` rather than `ApplicationRelativeTargetPath`. It is the manifest's target-path value; rooted values are rejected, but relative traversal such as `..` is still permitted pending containment validation.
- `IsFileExtensionMapped` is a Boolean rather than a nullable suffix because ClickOnce mapping adds only the fixed `.deploy` suffix.

I also updated the authoritative signing specification to require fail-fast deployment-reference validation before `ResolveFiles`. A deployment manifest must contain exactly one assembly dependency and no file references; that dependency must be `DeployManifest.EntryPoint` itself and must not be prerequisite, optional, or resource-fallback. Resource-fallback attributes are inspected in the preserved XML because `ManifestUtilities` neither models nor preserves them in its object model. Unsupported structures report the planned `--clickonce-signing-version 2 --no-update-clickonce-manifest` alternative.

Staging and signing remain outside this PR. The updated PR description records the additional integration requirements; they are not implemented by this resolver change.

src/Sign.Core/Tools/ClickOnce/IApplicationManifest.cs
```diff
@@ -2,12 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE.txt file in the project root for more information.
 
-using Microsoft.Build.Tasks.Deployment.ManifestUtilities;
-
 namespace Sign.Core
 {
     internal interface IApplicationManifest : IClickOnceManifest
```
kartheekp-ms, line 7: This interface has no contract. Do we need this interface?

src/Sign.Core/Tools/ClickOnce/ClickOnceApplicationPublishLayoutResolver.cs
```diff
@@ -0,0 +1,78 @@
...
+            ClickOncePayloadResolver payloadResolver)
+        {
+            ArgumentNullException.ThrowIfNull(manifestReader, nameof(manifestReader));
+            ArgumentNullException.ThrowIfNull(payloadResolver, nameof(payloadResolver));
+
+            _manifestReader = manifestReader;
+            _payloadResolver = payloadResolver;
+        }
+
+        internal bool TryResolve(FileInfo applicationManifestFile, out ResolvedClickOncePublishLayout? layout)
```
kartheekp-ms:
```suggestion
        internal bool TryResolve(FileInfo applicationManifestFile, [NotNullWhen(true)] out ResolvedClickOncePublishLayout? layout)
```

src/Sign.Core/Tools/ClickOnce/IApplicationManifest.cs
dtivel: Yes. Although it adds no application-specific members, it inherits the `IClickOnceManifest` contract and acts as the nominal application-manifest boundary. The reader returns it only after validating that the parsed model is an `ApplicationManifest`, and the resolved application and payload resolver require it so an `IDeployManifest` cannot be supplied in an application-only position. It also keeps those consumers substitutable in tests without coupling them to `ApplicationManifestAdapter`.

src/Sign.Core/Tools/ClickOnce/ClickOncePayloadResolver.cs
```diff
@@ -0,0 +1,445 @@
...
+                        throw new ClickOncePublishLayoutResolutionException(
+                            string.Format(
+                                CultureInfo.CurrentCulture,
+                                Resources.ClickOnceApplicationManifestResolveFailed,
+                                applicationManifestFile.FullName),
+                            diagnostics,
+                            exception);
+                    }
+
+                    AddDiagnostics(applicationManifest, diagnostics, ref diagnosticCount);
```
kartheekp-ms, line 125: For example, with `MapFileExtensions=true`, the manifest lists `mapped.dll`, but the published file is named `mapped.dll.deploy`. `ResolveFiles` reports an error for the missing `mapped.dll`, then our later lookup successfully finds `mapped.dll.deploy` (line no:153). This line keeps the earlier error in the successful result.

```
publish\
|-- App.application       (mapFileExtensions="true")
|-- App.exe.manifest      (references "mapped.dll")
`-- mapped.dll.deploy     (the actual published file)
```

Could we distinguish recovered lookup errors from actual failures? Otherwise, a caller logging `layout.Diagnostics` may report a missing file even though it was found.

src/Sign.Core/Tools/ClickOnce/ResolvedClickOncePayload.cs
```diff
@@ -0,0 +1,67 @@
...
+                source.Name,
+                $"{targetFileName}{DeploySuffix}",
+                StringComparison.OrdinalIgnoreCase))
+            {
+                isFileExtensionMapped = true;
+            }
+            else
+            {
+                throw new ArgumentException(
+                    "The source file name does not match the reference target path.",
```
kartheekp-ms, line 52: nit: Does sign CLI supports localized strings? If yes, should we localize all the error messages in this PR?

src/Sign.Core/Tools/ClickOnce/ClickOnceApplicationPublishLayoutResolver.cs
dtivel: Added `[NotNullWhen(true)]` to the `layout` parameter so the method's existing nullability contract is explicit. Thanks!

src/Sign.Core/Tools/ClickOnce/ResolvedClickOncePayload.cs
dtivel: Sign supports localization, and the user-facing operational errors added by this PR use localized resources. This message enforces an internal model invariant and is intended to indicate programmer misuse rather than an error shown to users, so I left it unlocalized.

FWIW, NuGet.Client similarly requires user-visible strings to use resources while retaining some hard-coded programmer-contract and internal-invariant exception messages, although it does not define a blanket policy based on exception type.

src/Sign.Core/Tools/ClickOnce/ClickOncePayloadResolver.cs
dtivel: Removed the preliminary `ApplicationManifest.ResolveFiles(...)` attempts because their narrower lookup semantics produced misleading errors when Sign CLI later resolved the reference through `.deploy` mapping or deployment-directory fallback. Sign CLI now reports an error only when all supported candidates fail. `ManifestUtilities.ResolveFiles(...)` will instead run against the finalized staged layout, where its warnings and errors are authoritative.

kartheekp-ms (approved):

PR #1054, opened 2026-08-11
Migrate tests from Moq to NSubstitute

kartheekp-ms:
Replace Moq across the test suite with NSubstitute so the repository can remove its legacy Moq dependency while preserving existing test behavior and coverage.

## Summary

- Replace centrally managed Moq packages with NSubstitute 6.0.0 and NSubstitute.Analyzers.CSharp 1.0.17.
- Keep package restore on the existing `dotnet-public` feed.
- Migrate routine setups, matchers, async returns, strict expectations, and received-call assertions to idiomatic NSubstitute.
- Replace protected `Dispose(bool)` verification with test-local recording spies.
- Use Azure SDK model/response factories and concrete substitutes for virtual SDK members.
- Remove the stale Moq reference from `Sign.TestInfrastructure` without changing production code.

## Notes for reviewers

`RSAKeyVault` exposes an internal constructor but grants Castle's `DynamicProxyGenAssembly2` friend access. NSubstitute works at runtime, but its analyzer cannot infer that access, so `NS2001` is suppressed only around the two proxy factory calls. The Key Vault tests exercise those proxies successfully.

## Validation

- `eng\common\build.cmd -restore -build -test`
- Build completed with 0 warnings and 0 errors; all test assemblies succeeded.

PR #1055, opened 2026-08-11
Do not sign ASP.NET Core static web assets in containers

### LINKED ISSUE #1045: Somehow distinguish .js files that are static web assets from those that are Windows Script Host scripts
**Is your feature request related to a problem? Please describe.**
We recently had a bug reported for a Blazor nuget package that was causing an integrity conflict and failure during development when that package was consumed (reference: https://github.com/IgniteUI/igniteui-blazor/issues/267)

Since by default, the `sign` tool treats NuGet packages as archives/containers and strives to apply code-siging practices to as many of the contents as possible, the problem is that for ASP.Net Core / Blazor-related NuGet packages that contain static web assets - `.js` files - the integrity of each file is recorded at build time in the `Microsoft.AspNetCore.StaticWebAssets.props` file.

However, if using the `sign` tool's default options, the `.js` asset files get changed and in development mode, the SHAs of the signed `.js` files mismatches those of the pre-recorded SHAs of the files prior to signing.

**Describe the solution you'd like**
I can imagine that considering .js files to be static web assets or WSH scripts is out of scope of the sign tool's logic, but may be it isn't. 
Either identify and omit .js files that are static web assets or at least throw a warning that such have been found and need to be excluded by either the --file-list option or outright using `--recurse-containers false`.

**Describe alternatives you've considered**
Even though it's useful to have all DLL files signed along with the containing NuGet packages, we resorted to using `--recurse-containers false` so that the .js web assets remain untouched. We are signing the DLL files prior to packaging the .nupkg file - just like when using the signtool.exe.

**Additional context**
sign tool version = 0.9.1-beta.26330.1+89c2a5d4a051014c243779d174bc3bb1fca62d7d

ChatchawanIllyes:
Fixes #1045

## Problem

Packages produced by the ASP.NET Core SDK (Blazor and Razor class libraries) carry their browser assets in a `staticwebassets` directory and record each asset's integrity hash at pack time in `build/Microsoft.AspNetCore.StaticWebAssets.props`.

`.js` is in the signable extension list because Windows Script Host scripts are `.js`, so recursing into such a package signs its `.js` assets. That changes the files, invalidates the recorded hashes, and consuming applications fail with an integrity conflict. The workaround reported in #1045 was `--recurse-containers false`, which gives up signing the package contents entirely.

## Change

When recursing into a container, `AggregatingSigner` now skips the files in a package's `staticwebassets` directory and logs a warning naming each skipped file.

Detection is convention-based and scoped to the package that declares the assets:

* A package is recognized by a `Microsoft.AspNetCore.StaticWebAsset*.props` file in its `build`, `buildTransitive`, or `buildMultiTargeting` directory. The wildcard also covers `Microsoft.AspNetCore.StaticWebAssetEndpoints.props` (.NET 9 and later).
* Only that package's own `staticwebassets` directory is skipped --- a `staticwebassets` directory elsewhere in the container is not.
* Files elsewhere in the package are unaffected, including `.js` files outside of `staticwebassets`, so genuine Windows Script Host scripts are still signed.

Filtering applies only when no file list was provided. A caller who explicitly listed files has already stated an intent, so static web assets can still be signed by listing them with `--file-list`.

The new `IStaticWebAssetFilter` works on the file paths already enumerated from the container, so it adds no `IContainer.GetFiles(...)` calls.

## Notes

* This changes default behavior: files that were previously signed inside static web asset packages are now skipped, with a warning. Signing a file that a browser integrity-checks is never the intent, so this seemed preferable to a warning-only change, but I am happy to switch to warning-only or to add an opt-out option if you would rather not change the default.
* One new resource string, with the `.xlf` files updated by the build.

## Testing

* New `StaticWebAssetFilterTests` covering the marker file locations and names, non-`.js` assets (`.wasm`, `.dll`), a `.js` file outside `staticwebassets`, a similarly named directory, case insensitivity, and the warning.
* New `AggregatingSignerTests` cases for a `.nupkg` container with and without a file list.
* I developed this on macOS, where `Sign.Core.Test` compiles but cannot run (the projects are `win-x64` and `Sign.TestInfrastructure` needs `mage.exe`). I verified the new and existing `AggregatingSigner` tests --- 55 in total --- by compiling the same sources into a scratch `net8.0` project and running them locally; they all pass. CI is the first Windows run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,21 @@ While the current version is limited to RSA and Azure Key Vault, it is desirable
 - ClickOnce `.application` and `.vsto` (via `Mage`). Notes below.
 - `.nupkg`
 
+## Static Web Assets
+
+Packages produced by the ASP.NET Core SDK (for example, Blazor and Razor class libraries) carry their browser
+assets in a `staticwebassets` directory and record each asset's integrity hash at pack time in
+`build/Microsoft.AspNetCore.StaticWebAssets.props`.  Signing such an asset --- most commonly a `.js` file, which is
+indistinguishable by extension from a Windows Script Host script --- changes the file and invalidates the recorded
+hash, which breaks consuming applications.
+
+When this tool recurses into a container and finds one of these `.props` files, it skips the assets in the
+package's `staticwebassets` directory and logs a warning for each skipped file.  Files elsewhere in the package,
+including `.js` files outside of `staticwebassets`, are unaffected.
+
+To sign static web assets anyway, list them explicitly using the `--file-list` option.  When a file list is
+provided, this tool signs whatever the list matches.
+
 ## ClickOnce
 There are a couple of possibilities for signing ClickOnce packages.
 
--- a/src/Sign.Core/DataFormatSigners/AggregatingSigner.cs
+++ b/src/Sign.Core/DataFormatSigners/AggregatingSigner.cs
@@ -13,6 +13,7 @@ namespace Sign.Core
         private readonly IFileMetadataService _fileMetadataService;
         private readonly IMatcherFactory _matcherFactory;
         private readonly IEnumerable<IDataFormatSigner> _signers;
+        private readonly IStaticWebAssetFilter _staticWebAssetFilter;
 
         // Dependency injection requires a public constructor.
         public AggregatingSigner(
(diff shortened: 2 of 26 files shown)


ChatchawanIllyes:
@dotnet-policy-service agree

kartheekp-ms:
@ChatchawanIllyes please rebase since we migrated tests from [Moq to NSubstitute](https://github.com/dotnet/sign/pull/1054).

PR #1058, opened 2026-08-16
Add safe ClickOnce publish-layout staging

dtivel:
Part of #1049.

## Summary

- Add dormant staging for `ResolvedClickOncePublishLayout` while preserving its typed deployment, application, payload, and adjacent-executable associations.
- Build an immutable staging plan and validate every destination before performing filesystem or manifest-reference changes.
- Reject rooted, traversing, escaping, colliding, and otherwise unrepresentable staging paths while preserving manifest-relative target paths.
- Copy the resolved publish layout into an isolated temporary directory and bind manifest references only to staged files.
- Temporarily remove only explicitly mapping-added `.deploy` suffixes while manifest file information is updated, then restore both filenames and references.
- Restore manifest references and clean up staged files during normal disposal and failure rollback.

## Design

Staging is mode-neutral. Signing and manifest-update policy remain responsibilities of the later application and deployment orchestration changes.

Planning and execution are separated: `ClickOnceStagingPlanBuilder` performs pure allocation and validation, while `ClickOncePublishLayoutStager` performs copying and reference binding. `ClickOnceStagingSession` owns the temporary staging lifecycle and mapped-suffix update scope.

## Behavior

No production signing behavior or CLI surface is changed. The staging implementation remains dormant until it is integrated by later ClickOnce signing changes.

## Testing

- Release restore, build, and all unit-test projects: 947 total, 946 passed, 1 skipped, 0 failed.
- Build completed with 0 warnings and 0 errors.
- Diff hygiene and stale staging-terminology checks passed.

### DIFF
--- a/src/Sign.Core/Tools/ClickOnce/ClickOnceStagedFile.cs
+++ b/src/Sign.Core/Tools/ClickOnce/ClickOnceStagedFile.cs
@@ -1,0 +1,36 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE.txt file in the project root for more information.
+
+using Microsoft.Build.Tasks.Deployment.ManifestUtilities;
+
+namespace Sign.Core
+{
+    internal sealed class ClickOnceStagedFile
+    {
+        internal ClickOnceStagedFile(
+            ClickOnceFileGraphEntry graphEntry,
+            FileInfo file,
+            FileInfo? manifestUpdateFile,
+            bool isUpdateInputOnly)
+        {
+            ArgumentNullException.ThrowIfNull(graphEntry, nameof(graphEntry));
+            ArgumentNullException.ThrowIfNull(file, nameof(file));
+
+            GraphEntry = graphEntry;
+            File = file;
+            ManifestUpdateFile = manifestUpdateFile;
+            IsUpdateInputOnly = isUpdateInputOnly;
+        }
+
+        internal ClickOnceFileGraphEntry GraphEntry { get; }
+        internal FileInfo Source => GraphEntry.Source;
+        internal FileInfo File { get; }
+        internal FileInfo? ManifestUpdateFile { get; }
+        internal string TargetPath => GraphEntry.TargetPath;
+        internal ClickOnceFileGraphEntryKind Kind => GraphEntry.Kind;
+        internal BaseReference? ManifestReference =>
+            GraphEntry.ManifestReference;
+        internal bool IsUpdateInputOnly { get; }
+    }
+}
--- a/src/Sign.Core/Resources.Designer.cs
+++ b/src/Sign.Core/Resources.Designer.cs
@@ -257,6 +257,51 @@ namespace Sign.Core {
                 return ResourceManager.GetString("ClickOnceSignatureProviderSigning", resourceCulture);
             }
         }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Failed to prepare file &apos;{0}&apos; for the signing operation..
+        /// </summary>
+        internal static string ClickOnceStagingCopyFailed {
+            get {
+                return ResourceManager.GetString("ClickOnceStagingCopyFailed", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Source files &apos;{0}&apos; and &apos;{1}&apos; have conflicting target paths &apos;{2}&apos; and &apos;{3}&apos;..
+        /// </summary>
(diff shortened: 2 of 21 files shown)


src/Sign.Core/Tools/ClickOnce/ClickOnceStagingPlanBuilder.cs
```diff
@@ -0,0 +1,699 @@
...
+        private static bool IsWindowsDriveQualified(string path)
+        {
+            return path.Length >= 2 &&
+                path[1] == ':' &&
+                char.IsAsciiLetter(path[0]);
+        }
+
+        private static bool IsWindowsReservedFileName(string segment)
+        {
+            string fileName = Path.GetFileNameWithoutExtension(segment);
```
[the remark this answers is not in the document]
dtivel: Updated the validation to compare the portion of each path segment before its first `.`. Added a regression case for `NUL.tar.gz\payload.dll` that verifies the layout is rejected before copying.

src/Sign.Core/Tools/ClickOnce/ClickOnceStagedFile.cs
kartheekp-ms, line 1: I am curious to understand how is this class different from `ClickOnceStagingPlanEntry.cs`. They have the same properties.

src/Sign.Core/Tools/ClickOnce/ClickOnceStagingSession.cs
```diff
@@ -0,0 +1,389 @@
...
+                            exception));
+                }
+            }
+
+            _manifestFileInfoUpdateActive =
+                _mappedFiles.Any(file => file.IsAtUpdatePath);
+            ThrowRestoreFailures(exceptions);
+        }
+
+        private IReadOnlyList<Exception> RestoreMappedFiles(
```
kartheekp-ms, line 217: This method and `EndManifestFileInfoUpdate` have similar logic in the foreach loop. To avoid maintainability issues in the future, I suggest extract common logic to another method.

src/Sign.Core/Tools/ClickOnce/ClickOncePublishLayoutStager.cs
```diff
@@ -0,0 +1,159 @@
...
+                    if (!copiedDestinations.Add(
+                        entry.Destination.FullName))
+                    {
+                        continue;
+                    }
+
+                    entry.Destination.Directory!.Create();
+                    entry.Source.CopyTo(
+                        entry.Destination.FullName,
+                        overwrite: false);
```
kartheekp-ms, line 107: `FileInfo.CopyTo` preserves the source file’s read-only attribute. For a read-only payload, normal disposal and copy-failure rollback can therefore leave staged files behind: `DirectoryService.Delete` logs and suppresses the deletion failure. [.NET documentation](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.delete?view=net-8.0#system-io-directoryinfo-delete(system-boolean)) explicitly lists `UnauthorizedAccessException` when the directory contains a read-only file. Can we clear `FileAttributes.ReadOnly` on the staged copy after copying, without changing the source’s attributes?

src/Sign.Core/Tools/ClickOnce/ClickOnceStagedFile.cs
dtivel: `ClickOnceStagingPlanEntry` represents validated intent before filesystem changes, while `ClickOnceStagedFile` represents the successfully materialized result exposed by `ClickOnceStagingSession`. The identical shape is intentional type-state separation so planned destinations cannot be treated as existing staged files.

src/Sign.Core/Tools/ClickOnce/ClickOnceStagingSession.cs
dtivel: Extracted the shared per-file restoration logic into `RestoreMappedFile`. `EndManifestFileInfoUpdate` and failed-begin rollback retain their separate iteration and exception-handling behavior.

src/Sign.Core/Tools/ClickOnce/ClickOncePublishLayoutStager.cs
dtivel: I am deferring read-only handling to the integration and copy-back work. Clearing the attribute here would establish only part of the policy without defining whether original attributes must be restored when signed files are copied back. Rejecting read-only sources here would also be premature because some staged files may only need to be read or copied, not modified. Existing Sign CLI flows generally assume writable files and do not preserve attributes explicitly, so this PR will not introduce a ClickOnce-specific partial policy.

kartheekp-ms (approved):

kartheekp-ms (approved):

PR #1066, opened 2026-09-06
Add signing operation coordination infrastructure

dtivel:
## Summary

- Add canonical signing-source identities for physical files and recursively nested container entries.
- Add invocation-scoped coordination so duplicate source claims share one owner operation, result, or failure.
- Retain immutable snapshots for successful operations and support atomic materialization into multiple layouts.
- Add disposal-aware snapshot lifetime management and direct self-await protection.
- Add focused coverage for identity semantics, concurrency, shared failures, reentrancy, materialization, and cleanup.

## Behavior

This infrastructure remains dormant. It is not registered with dependency injection or integrated with dispatch, signing formats, containers, or CLI behavior. Activation is deferred to a later PR in the ClickOnce implementation sequence.

## Testing

- `dotnet build .\sign.sln --no-restore`
- `dotnet test .\sign.sln --no-build --no-restore`

993 tests passed and 1 test was skipped.


test/Sign.Core.Test/TestInfrastructure/TestDirectory.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        internal string FullPath { get; }
+
+        internal FileInfo CreateFile(
+            string relativePath,
+            string? contents = null)
+        {
+            string path = Path.Combine(FullPath, relativePath);
+            Directory.CreateDirectory(Path.GetDirectoryName(path)!);
+            File.WriteAllText(path: path, contents: contents);
```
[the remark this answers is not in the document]
dtivel: Fixed. `CreateFile` now rejects rooted paths, normalizes the combined path, and verifies it cannot escape the owned test directory before creating anything. Added focused rooted-path and `..` traversal tests.

src/Sign.Core/SigningOperationCoordinator.cs
```diff
@@ -0,0 +1,213 @@
...
+                UnauthorizedAccessException)
+            {
+                Exception primaryException =
+                    exception is FileNotFoundException ||
+                    exception is DirectoryNotFoundException &&
+                        !File.Exists(artifact.FullName)
+                        ? new InvalidOperationException(
+                            message: "The signed artifact is unavailable.",
+                            innerException: exception)
+                        : exception;
```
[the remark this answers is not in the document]
dtivel: Fixed. Extracted the classification into an `artifactIsUnavailable` boolean and explicitly grouped the `DirectoryNotFoundException` source-existence check, preserving the existing behavior while making the precedence clear.

src/Sign.Core/SigningOperationCoordinator.cs
```diff
@@ -0,0 +1,215 @@
...
+                // same-identity claim from the owner's context is rejected as
+                // self-await; delegates must not leave such work unawaited.
+                // Planning must prevent cycles between distinct operations.
+                if (ReferenceEquals(s_executingState.Value, this))
+                {
+                    throw new InvalidOperationException(
+                        message: "A signing operation cannot await itself.");
+                }
+
+                if (Interlocked.Exchange(ref _started, value: 1) == 0)
```
kartheekp-ms, line 147: Could we simplify this by making the caller whose candidate is inserted by `GetOrAdd` responsible for starting the operation?

In `ExecuteAsync`, `ReferenceEquals(state, candidate)` tells us whether this caller inserted the shared state. Only that caller would start the operation; callers that find an existing state would simply wait on its completion task. With that ownership rule, the separate `_started` flag and `Interlocked.Exchange` would no longer be necessary.

We could also combine `Start` and `CompleteAsync` into one `RunAsync(operation, CreateSnapshot)` method, preserving the shared `TaskCompletionSource`. Passing the callbacks into that method instead of storing them in `OperationState` would avoid retaining the signing delegate and its captured objects for the rest of the invocation after completion.

src/Sign.Core/SigningSourceIdentity.cs
```diff
@@ -0,0 +1,161 @@
...
+{
+    internal sealed class SigningSourceIdentity :
+        IEquatable<SigningSourceIdentity>
+    {
+        private readonly IdentityKind _kind;
+        private readonly string? _path;
+        private readonly SigningSourceIdentity? _parent;
+
+        private SigningSourceIdentity(
+            IdentityKind kind,
```
kartheekp-ms, line 15: Do we need the `kind` parameter if we can derive the identity kind from `parent`? My understanding is that a null `parent` represents a physical file, while a non-null `parent` represents an entry inside a container.

kartheekp-ms (dismissed):

src/Sign.Core/SigningOperationCoordinator.cs
dtivel: Good suggestion. Updated in `62145d8`. The `GetOrAdd` insertion winner now starts the operation through a combined `RunAsync` method. I removed `_started`/`Interlocked.Exchange` and stopped storing the operation and snapshot callbacks in `OperationState`, so their captures are released after execution. Delegate invocation remains inline before the first await, with the existing self-await detection and shared completion semantics preserved. I also added regression coverage for synchronously thrown operation exceptions.

src/Sign.Core/SigningSourceIdentity.cs
dtivel: Agreed. Updated in `62145d8`. I removed `IdentityKind` and the `kind` constructor parameter; parent presence now structurally distinguishes container entries from physical files. This also allowed `_path` to become non-nullable and removed the unreachable switch defaults and null-forgiving operators. I added coverage for identical entry paths under different parents and for physical-file versus container-entry inequality.

src/Sign.Core/SigningOperationCoordinator.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        public void Dispose()
+        {
+            _snapshotLifetime.Dispose();
+        }
+
+        private SigningOperationResult CreateSnapshot(FileInfo artifact)
+        {
+            using IDisposable lease =
+                _snapshotLifetime.EnterOperation();
```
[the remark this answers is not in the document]
dtivel: Fixed in `784e94f`. The coordinator now acquires an operation lease before publishing a candidate into the shared dictionary. Losing candidates release their leases immediately, while the winning lease remains active through delegate execution and snapshot publication, so disposal defers snapshot-directory cleanup without accepting new work. `CreateSnapshot` no longer attempts a late lease acquisition. Added a deterministic test that disposes the coordinator while an owner is blocked, verifies cleanup is deferred, then verifies the owner completes successfully and cleanup occurs after its lease is released.

kartheekp-ms (approved):

PR #1046, opened 2026-07-27
Add Azure Key Vault certificate version option

Guflly:
Adds `--azure-key-vault-certificate-version` / `-kvcv` and uses the selected certificate and key version.

Resolves #1005

PR #1074, opened 2026-09-15
Add Azure Key Vault certificate version option

### LINKED ISSUE #1005: Add certificate version option (-kvcv) when using azure-key-vault
Currently, it appears there is no option to specify a certificate version when using key vault.
AzureSignTool allows for this via -kvcv 

Add certificate version option (-kvcv) when using azure-key-vault

dtivel:
Adds `--azure-key-vault-certificate-version` / `-kvcv` to select a specific Azure Key Vault certificate version.

The implementation:

- Validates versions using the publicly documented 32-character format.
- Retrieves the requested certificate version without falling back to latest.
- Uses the certificate's authoritative backing `KeyId`, since certificate and key versions can differ.
- Validates and normalizes missing or malformed backing key identifiers.
- Caches the selected certificate and cryptography client.

Based on #1046 by @Guflly, preserving the original commit authorship.

Resolves #1005

## Testing

- Full test suite: 813 passed, 1 skipped.
- Focused certificate-version and invalid-KeyId tests: 16 passed.
- Baseline/candidate release matrix: 12 signing scenarios passed across NuGet, ClickOnce, and VSTO using local certificates and Azure Key Vault.
- Verified explicit current-version signing and rejection of invalid-length values.


src/Sign.Cli/AzureKeyVaultCommand.cs
```diff
@@ -135,5 +151,17 @@ internal AzureKeyVaultCommand(CodeCommand codeCommand, IServiceProviderFactory s
 
             return uri;
         }
+
+        private static string? ParseCertificateVersion(ArgumentResult result)
+        {
+            if (result.Tokens.Count != 1 || result.Tokens[0].Value.Length != 32)
```
kartheekp-ms, line 147: Should we also check the certificate version for any invalid characters in addition to the length check?

src/Sign.Cli/AzureKeyVaultCommand.cs
```diff
@@ -93,25 +101,33 @@ internal AzureKeyVaultCommand(CodeCommand codeCommand, IServiceProviderFactory s
...
                 {
                     services.AddAzureClients(builder =>
                     {
                         builder.AddCertificateClient(certId.VaultUri);
-                        builder.AddCryptographyClient(keyUri);
                         builder.UseCredential(credential);
                         builder.ConfigureDefaults(options => options.Retry.Mode = RetryMode.Exponential);
                     });
 
+                    services.AddSingleton<Func<Uri, CryptographyClient>>(_ =>
```
kartheekp-ms: Could we register a `KeyClient` with `AddAzureClients` and use its [`GetCryptographyClient(keyName, keyVersion)`](https://learn.microsoft.com/dotnet/api/azure.security.keyvault.keys.keyclient.getcryptographyclient?view=azure-dotnet) method here? That would reuse the shared client options and avoid duplicating configuration in this custom factory.

Copilot suggested following pseudo code will work in this scenario.

```cs
// AzureKeyVaultCommand: register both clients for the same vault.
services.AddAzureClients(builder =>
{
    builder.AddCertificateClient(certId.VaultUri);
    builder.AddKeyClient(certId.VaultUri);

    // The cryptography client created through KeyClient will inherit
    // its authentication pipeline and configured options.
    builder.UseCredential(credential);
    builder.ConfigureDefaults(options =>
        options.Retry.Mode = RetryMode.Exponential);
});

// KeyVaultService: replace the custom factory dependency with KeyClient.
// Assign this field from the constructor-injected client.
private readonly KeyClient _keyClient;

// When registering KeyVaultService, pass this resolved client
// to its constructor instead of Func<Uri, CryptographyClient>.
serviceProvider.GetRequiredService<KeyClient>();

// During the existing backing-key validation, retain the parsed identifier.
// This call must remain guarded: reject parse failures and preserve the
// existing checks for the configured vault, object type, and key version.
KeyVaultKeyIdentifier.TryCreate(keyId, out var keyIdentifier);

// Store the validated identifier in a new CertificateInfo.KeyIdentifier
// property, alongside the cached certificate.

// GetRsaAsync: keep the existing initialization lock and client caching.
// Use the backing key's name/version, NOT the certificate's version.
_cryptographyClient ??=
    _keyClient.GetCryptographyClient(
        certificateInfo.KeyIdentifier.Name,
        certificateInfo.KeyIdentifier.Version);
```

kartheekp-ms (dismissed):

src/Sign.Cli/AzureKeyVaultCommand.cs
dtivel: I resisted the temptation. I looked into this before adding the length check. The [Key Vault documentation](https://learn.microsoft.com/azure/key-vault/general/about-keys-secrets-certificates#object-identifiers) defines an object version as a system-generated, 32-character string, but does not document any restriction on its characters. The .NET SDK documentation and XML comments likewise specify no character set and leave validation of the version to the service. Since hexadecimal-looking examples are not a documented contract, I intentionally limited client-side validation to the documented length.

src/Sign.Cli/AzureKeyVaultCommand.cs
dtivel: Thanks. Updated this to register `KeyClient` through `AddAzureClients` and call `GetCryptographyClient` with the name and version parsed from the certificate response's authoritative `KeyId`. I also added coverage for differing certificate/backing-key identifiers and one-time client creation.

kartheekp-ms (approved):

PR #1076, opened 2026-09-20
Apply --file-list filtering to standalone input files

minahilatiqdev:
Fixes #990

## Problem
`--file-list` was only applied to files discovered inside containers (zip/appx/nupkg) and ClickOnce deployments. Standalone files passed directly as positional arguments were not filtered, so the option was silently ignored for the most common use case.

## Change
Extended the existing file-list filtering (via `FileListReader`/`IFileMatcher`) so it is applied consistently to standalone input files as well, while leaving container and ClickOnce deployment inputs unfiltered at the top level (their contents continue to be filtered separately, as before).

## Tests
Added 6 new tests covering:
- Standalone files are filtered according to include patterns
- Exclusion-only file lists do not implicitly include everything
- Include-all + exclusion patterns work together
- Container inputs are not filtered at the top level (regression check)
- File-list patterns matching files in subdirectories
- Mixed container + standalone file inputs are handled correctly

Full test suite (427 tests) passes with no regressions.


minahilatiqdev:
@dotnet-policy-service agree

minahilatiqdev:
Hi @kartheekp-ms, just checking in — this PR is ready for review whenever you have a chance (CLA signed, CI checks passing). Let me know if any changes are needed. Thanks!

